PR #33, opened 2025-12-07
Fix upfile parsing bounds and file count guard

james-coder:
- Bound trim_right to max_len and force NUL to avoid reading past fixed-size buffers when headers lack a terminator.
  - Guard against unreasonable header.files and use a uint16_t loop counter to prevent wrap/hangs on malformed UPFILEs.


james-coder:
I should note that the change is simplistic enough to be easily understood, but I don't have a U1 yet (April 2026🤞) so it wasn't tested on an actual firmware build/flash.

PR #34, opened 2025-12-07
Add watchdog restart for v4l2-mpp capture services

james-coder:
- Add a lightweight monitor loop to S99v4l2-mpp-{mipi,usb} that restarts capture-*-mpp if it exits, preventing the camera stream from staying black after a crash or timeout.
  - Use pidfiles for both capture and monitor; ensure only one monitor runs and clean it up on stop.
  - No other behavior changes to the streaming stack; just keeps the producer alive.

## Test
On the printer (or a device with the firmware unpacked/running), you can simulate a crash and watch it restart:

Kill the producer; monitor should restart it within 5s
```  
kill -9 "$(cat /var/run/capture-mipi-mpp.pid)"   # or capture-usb-mpp.pid
```

Check it came back
```
sleep 6
ps | grep capture-mipi-mpp   # or capture-usb-mpp
```

Also verify monitor cleanup on stop:
```
/etc/init.d/S99v4l2-mpp-mipi stop
pgrep -f capture-mpp-monitor  # should be empty
```
And ensure the monitor doesn’t multiply:
```
/etc/init.d/S99v4l2-mpp-mipi start
/etc/init.d/S99v4l2-mpp-mipi start
pgrep -f capture-mpp-monitor  # should show only one monitor pid
```


paxx12:
@james-coder I made the following https://github.com/paxx12/v4l2-mpp/pull/17 and https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/35 as this gives more control in the future on the behavior.

PR #46, opened 2025-12-13
Add IPv6 workaround for SnOrca camera

horzadome:
- force moonraker to prioritize ipv4 even if ipv6 is available
  - this only applies to MQTT clients (only SnOrca?), not all moonraker clients (fluidd, homeassistant other normal http clients should be unaffected)
  - only applies in dual-stack (ipv4+ipv6) environments
    - Camera works just fine in ipv4-only environments so no fix needed
    - In ipv6-only environments it can't be worked around at all because of SnOrca

SnOrca's ipv6 is currently broken and unsupported.
I tried to isolate and minimize this horrible workaround to not break ipv6 for clients that work just fine.
Intentionally called overlay "ipv6-workarounds-for-snorca" because I expect we'll have more than this one and that we'll want to easily find and remove them all when SnOrca's ipv6 gets fixed.


horzadome:
After talking with @paxx12 agreed that #47 is the way to go instead.

PR #54, opened 2025-12-17
Add webcam stream type selection

horzadome:
- add template for multiple internal webcam stream types
- keep the *.cfg.disabled model, align all camera configs with it
- enabling user webcam config disables default "case" config

Note: replaces PR #52 because I messed up reflog in that one.

### DIFF
--- a/overlays/camera-v4l2-mpp/root/home/lava/origin_printer_data/config/moonraker/09_user_camera.cfg.disabled
+++ b/overlays/camera-v4l2-mpp/root/home/lava/origin_printer_data/config/moonraker/09_user_camera.cfg.disabled
@@ -1,0 +1,41 @@
+# To customize camera streams:
+# 1. Rename this file to 09_user_camera.cfg using Fluidd web interface or SSH shell.
+# 2. Enable stream types that you want by changing "enabled" to true.
+#    Enabling multiple stream types will add multiple cameras to Fluidd and other clients.
+# 3. Feel free to adjust other parameters as needed.
+
+
+# Default Internal camera using WebRTC
+# This webcam must be named "case" to override default settings from 02_default_internal_camera.cfg
+# Enabling this camera and modifying its settings here will override default settings from 02_default_internal_camera.cfg
+[webcam case]
+enabled: false
+service: webrtc-camerastreamer
+stream_url: /webcam/webrtc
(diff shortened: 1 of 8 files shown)


horzadome:
I tested this and user config survives firmware updates unless we rename the default webcam; then it gets re-enabled until user renames it in user config file

overlays/camera-v4l2-mpp/root/home/lava/origin_printer_data/config/moonraker/09_user_camera.cfg.disabled
```diff
@@ -0,0 +1,41 @@
...
+# 1. Rename this file to 09_user_camera.cfg using Fluidd web interface or SSH shell.
+# 2. Enable stream types that you want by changing "enabled" to true.
+#    Enabling multiple stream types will add multiple cameras to Fluidd and other clients.
+# 3. Feel free to adjust other parameters as needed.
+
+
+# Default Internal camera using WebRTC
+# This webcam must be named "case" to override default settings from 02_default_internal_camera.cfg
+# Enabling this camera and modifying its settings here will override default settings from 02_default_internal_camera.cfg
+[webcam case]
```
paxx12: I do like this format the most. Maybe we should also apply this for USB.

Also, why we would like people to use MJPEG variants? Maybe we should ask people to select just one of the streams? In all cases the `snapshot_url:` is the same. So, maybe we could have that way:

```
# Use WebRTC (the highest performance)
[webcam case]
enabled: true
service: webrtc-camerastreamer
stream_url: /webcam/webrtc
snapshot_url: /webcam/snapshot.jpg

# Use H264 or MJPEG depending on client compatibility
[webcam h264]
enabled: false
service: iframe
stream_url: /webcam/player
snapshot_url: /webcam/snapshot.jpg
aspect_ratio: 16:9
target_fps: 15

# Use MJPEG (the best compatibility, but most resource intensive)
[webcam mjpeg]
enabled: false
service: mjpegstreamer-adaptive
snapshot_url: /webcam/snapshot.jpg
target_fps: 15
target_fps_idle: 1
```

We will need to test this also on Mainsail if it works.

paxx12:
Replaced by https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/134/files

PR #56, opened 2025-12-17
Add curl

horzadome:
- add system-utils overlay to "extended" profile for things like curl
- install curl to /usr/local/bin
- add curl to dev Dockerfile so container can use curl to download curl

Reasoning:
There is no curl in stock firmware.
We would like to have curl available so that we can e.g. interact with Moonraker API. Specifically, in future PR I'll make adding/removing USB camera to Moonraker on hotplug so that users don't have to manually add it to Moonraker/Fluidd.

Curl is installed from latest binary release of https://github.com/stunnel/static-curl (Travis Lee). This download location is featured at https://curl.se/download.html so it's reliable enough for us to use.

### DIFF
--- a/overlays/firmware-extended/20-system-utils/scripts/curl_install.sh
+++ b/overlays/firmware-extended/20-system-utils/scripts/curl_install.sh
@@ -40,3 +40,4 @@ if [[ ! -x "$1/usr/local/bin/curl" ]]; then
 fi
 
 echo ">> curl installed to $1/usr/local/bin"
+


overlays/system-utils/scripts/curl_install.sh
```diff
@@ -0,0 +1,41 @@
...
+if [[ $# -ne 1 ]]; then
+  echo "Usage: $0 <rootfs-dir>"
+  exit 1
+fi
+
+set -eo pipefail
+
+printf ">> Installing latest curl from precompiled tarball\n"
+
+REPO_URL=https://github.com/stunnel/static-curl
```
paxx12: Can we similar to other put a static URL, with SHA256SUM?

`wget` it into `tmp/` if not existing, unpack this.

The reason for this is:

- At some point we tested a specific version to behave as expected. This is guarded by the version and sha256 so we have a clear SBOM and prevent someone changing our depdendency for whatever reasons. In the end us shipping the firmware to user machines should ensure that we do not introduce malicious binaries.
- Curl is not supposed to change very frequently, in general it is better to have some dependency bot to update the dependency as part of PR instead of pulling from latest. I think we could probably have copolit periodically look at those things.

overlays/system-utils/scripts/curl_install.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/env bash
```
paxx12: To be removed

overlays/firmware-extended/20-system-utils/scripts/curl_install.sh
```diff
@@ -0,0 +1,42 @@
...
+echo ">> Verifying /usr/local/bin/curl checksum..."
+echo "$BIN_SHA256  $1/usr/local/bin/curl" | sha256sum --check --status
+
+echo ">> Verifying if curl is executable..."
+if [[ ! -x "$1/usr/local/bin/curl" ]]; then
+  echo "[!] /usr/local/bin/curl is not executable."
+  exit 1
+fi
+
+echo ">> curl installed to $1/usr/local/bin"
```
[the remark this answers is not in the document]
horzadome: I better stop touching computers for today because even my editor no longer does what I want it to :D

PR #66, opened 2025-12-19
Draft: Cameras using moonraker API instead of static config

horzadome:
- add cameras using Moonraker API instead of static config files
- cameras added this way can be edited by user in Fluidd
- USB camera left disabled by default; user can enable it from Fluidd


paxx12:
@horzadome I'm not convinced by having dynamic hotplug to inject config into moonraker. It is nice, but I'm also unsure if this is needed at the moment. However, the idea is clever and we could consider this in the future. As of now I do rather prefer static configs as a way to enable things.

PR #117, opened 2026-01-09
Read alpha and multicolor info from NFC

macdylan:
Currently printer screen do not support displaying alpha channel, but SNOrca can. And SN protocol support multiple colors, which may soon be displayed on the screen.

```
{
    "protocol": "openspool",
    "version": "1.0",
    "brand": "Protopasta",
    "type": "PLA",
    "subtype": "",
    "color_hex": "ffffff00",
    "multi_color_hexes":"d2629c,EE2679,9e93f3",
    "min_temp": "190",
    "max_temp": "220",
    "bed_max_temp": "60"
}
```

### DIFF
--- a/overlays/firmware-extended/13-rfid-support/root/home/lava/klipper/klippy/extras/filament_protocol_ndef.py
+++ b/overlays/firmware-extended/13-rfid-support/root/home/lava/klipper/klippy/extras/filament_protocol_ndef.py
@@ -26,29 +26,6 @@ def xxd_dump(data, max_lines=16):
 
     return '\n'.join(lines)
 
-def parse_hex_color(color_hex):
-    if color_hex.startswith('#'):
-        color_hex = color_hex[1:]
-
-    if len(color_hex) not in [6, 8]:
-        raise ValueError(f"Invalid color: {color_hex}")
-
-    r = color_hex[0:2]
-    g = color_hex[2:4]
-    b = color_hex[4:6]
-    alpha = color_hex[6:8] if len(color_hex) == 8 else "ff"
-
-    try:
-        rgb = int(r + g + b, 16)
-        alpha = int(alpha, 16)
-    except ValueError:
-        rgb = 0xFFFFFF
-        alpha = 0xFF
-
-    argb = alpha << 24 | rgb
-
-    return rgb, alpha, argb
-
 def ndef_parse(data_buf):
     if None == data_buf or isinstance(data_buf, (list, bytes, bytearray)) == False:
         return NDEF_PARAMETER_ERR, []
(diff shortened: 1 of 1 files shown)


paxx12:
@macdylan 

I think I would prefer that we do:

```json
    "alpha": "#FF",
    "additional_color_hexes":["#d2629c","#EE2679","#9e93f3"],
```

Also, ensure that those non-standard tags are documented.

paxx12:
@macdylan 

Please do not change the semantics of `color_hex`. It should be always `RGB`, not `RGBA`. Please document all those additions in the `rfid_support.md`.

macdylan:
I have some translucent PETG materials, and their transparency can be correctly displayed in SNOrca. This is very helpful for me in choosing the right materials, especially white and clear white. However, the screen does not support displaying them yet.

And I have already contacted the Snapmaker devteam, and this request has been scheduled.

macdylan:
I think compatibility is more important than "field name". That is an "OPEN" protocol, and you don't know what kind of data users will write. For example, I might export JSON data from Spoolman to OpenSpool, but the color information in Spoolman may include `alpha`. You rejected this kind of compatibility, so I closed it because it deviates from my actual needs and original idea.

paxx12:
@macdylan 

> You rejected this kind of compatibility, so I closed it because it deviates from my actual needs and original idea.

The changes introduced needs to be additive to the protocol. **We cannot change the semantics of the fields as this breaks the clients that do read the tags.** The project supporting RGBA on `color_hex` will automatically create drift to users using incompatible applications as some applications will be able to read such tags some do not. We cannot make this sometimes work or not, because you used incompatible client. This is the reason why I rejected the change you are making to `color_hex`. This will be bad user-experience in the end, and additional support requests.

The latest change do allow to set `alpha` and other fields in a way that does not break other existing clients - by leveraging additional fields.

The Spoolman has no knowledge of OpenSpool. The Spoolman does not know what format the OpenSpool is accepting, as such usually you have converter that transforms what is Spoolman providing to what is the OpenSpool expecting, where the fields, names, content are re-mapped.

The end result is that we allow to provide `alpha` and `additional_color_hexes`, but you need to set this as an additional field and everything works consistent for everyone else.

patbearnl:
i have already 2 application made for converting spoolman and 3d filament profiles. one is a android app that writes openspool tags. and the other is a windows app that converts the same jsons/cvs to snorca profiles. it would be easy to add a toggle to add or discard the alpha channel from the hex. it already uses paxx12 suggestion for additional and alpha strings. it now is just a thing if those string are getting tunneld to snorca. if so then it wouldnt be to hard to add alpha en multicolor to snorca. but paxx12 is right. you cant just change the openspool protocol. its made to work in a way and you can add to that way but not change existing.

macdylan:
> if so then it wouldnt be to hard to add alpha en multicolor to snorca

In fact, RGBA has already been used in SNOrca.

patbearnl:
and it wants it as #ffffffaa ?

patbearnl:
@paxx12  is it possible to add the alpha string to the opentag like you have done alpha="" and then make a little something in the parser that joins the alpha value to the rgb value so snorca recieves rgb="ffffff00" since snorca can understand that value. this no code breaking maner and makes thing simple. openspool protocol on bambulabs ignore the alpha openspool stays clean. snorca stays clean and the u1 is a feature richer?? maybe this is a good middle ground.

PR #146, opened 2026-01-16
Night Mode

egette:
Apply TMC2240 autotune registers only during Night Mode, extend profile & scheduler

- TMC2240 autotune registers for stepper_x/stepper_y are now set only when Night Mode is active and restored on exit (no longer persistent in config).
- New option `profile_tmc_autotune_enabled` in the [night_mode] profile (default: true) controls this behavior.
- Profiles for speed, acceleration, jerk, fan, stepper current, and TMC autotune are loaded from extended.cfg by the scheduler, validated, and passed to the macros.
- Safety floors for all profile properties (e.g. speed ≥ 70%, accel ≥ 5000, fan 50–100%, current 0.1–2.5A).
- Scheduler script and init script added/updated to manage profiles and time windows automatically.
- Macros reliably save and restore all values (including stepper current and TMC registers).
- Documentation and example configurations updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ Get the latest pre-built firmware from [Releases](https://github.com/paxx12/Snap
 
 See [User Documentation](https://snapmakeru1-extended-firmware.pages.dev/) for features, installation instructions, and usage guides.
 
+- Quiet profile: [Silent Mode Macros](docs/silent_mode.md) – `SILENT_ON`/`SILENT_OFF` macros and how to auto-apply them before prints or at boot
+
 ## Building from Source
 
 See [Building from Source](docs/development.md) for instructions on building custom firmware using Docker.
--- a/docs/silent_mode.md
+++ b/docs/silent_mode.md
@@ -1,0 +1,97 @@
+---
+title: Silent Mode Macros
+---
+
+# Silent Mode Macros
+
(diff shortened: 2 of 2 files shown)


paxx12:
This seems to be heavily AI generated given the way that data is processed and how it is implemented. It would be good that such content is properly described to be, as this indicates that those changes are very unlikely to be tested.

---

The idea is inspired by Prusa Silent mode. This seems that can be implemented with just a `[gcode_macro SILENT_ON]` and `[gcode_macro SILENT_OFF]`.

---

It also seems that people's general preference is to enable Silent forever or Normal forever. I don't think there's a need to change such things dynamically.

---

So, the idea seems cool, but at the same time this brings crazy amount of complex logic that does have affect on printing quality, surface quality, etc.:

- It changes TMC driver values by default
- Tries to be clever to apply the speed limits, but does not prevent the running gcode to change those
- Introduces `scheduler.py` that seems to be basically `cron`
- For some reason we try to expose all those settings via `extended.cfg` where-as general approach is to keep such things in a `klipper.cfg` files
- Does not understand the printer state, and what changes were made to the printer

It seems that it is better to have those configured per-print:

- You really want to make to use consistent settings throughout the given print as it has impact on finish quality of the prints.

But, if there's a need for schedule:

- You want to use the `cron` and send the command to klipper to apply different limits at a given time.
- The proposed `gcode_macros` could contain variables that would set machine limits, and would have to reset.

paxx12:
@egette I do not plan to accept this PR in the current form. I do plan to adopt TMC Auto Tune parameters, but this is rather to be always enabled or always disabled based on user choice.

PR #156, opened 2026-01-19
Add support for writing RFID tags

Lemmons:
Web ui and apis runs as part of moonraker, accessible at http://{u1-hostname/ip}/rfid. Includes all rfid fields supported in v1.0.0.  Also adds new new GCODE commands to support the web UI functionality.  The web UI also contains an export and import tag functionality for ease of cloning/backing up tags, etc.

# UI General Look
image
image
image
image

# Empty and Missing Tag support
Properly shows when a tag is empty or has an error vs missing.
image
image

# Mobile Browser Support
Also renders and works well on a mobile browser.
RFID Tag Manager - Snapmaker U1

# Snapmaker Native Tags
Native Snapmaker filament tags are support for reading, but not writing or erasing.  The UI prevents submitting, and the api also rejects requests to modify them.


paxx12:
That is crazy :) I'm currently on holidays will review this in a few days, as there's a number of changes.

How you also intend to guide user to write both sides of spool?

Lemmons:
> That is crazy :) I'm currently on holidays will review this in a few days, as there's a number of changes.
> 
> How you also intend to guide user to write both sides of spool?

I have been thinking about that.  Currently I'm thinking to do the "simple" thing and have a prompt for the user to confirm/dismiss any time they write a tag saying they should update the other side as well.  That will also likely necessitate adding a "clone" functionality of some sort.  That being said, I'd prefer to add those in as a separate pr, and focus on getting these changes in (with your feedback on what is here now), as the pr already is pretty large.

overlays/firmware-extended/13-rfid-support/patches/04-add-gcode-write-commands.patch
```diff
@@ -0,0 +1,336 @@
+   Including: home/lava/klipper/klippy/extras/filament_detect.py
+--- rootfs.original/home/lava/klipper/klippy/extras/filament_detect.py	2026-01-19 20:46:51.443456566 +0000
++++ rootfs/home/lava/klipper/klippy/extras/filament_detect.py	2026-01-19 20:28:56.743188393 +0000
+@@ -17,6 +17,18 @@
+ FILAMENT_DT_CHANNEL_NUMS                        = 4
+ FILAMENT_DT_CONFIG_FILE                         = "filament_detect.json"
+ 
++# Material density defaults (g/cm³) for tag writing
++MATERIAL_DENSITIES = {
```
unlucio: what about mathicing Orca's material list here:
https://github.com/OrcaSlicer/OrcaSlicer/blob/b93112c67431d2cb962f2b6b7ad21fb13989119e/src/libslic3r/MaterialType.cpp#L92

But in all honesty, considering the variety of offerings on the market, locking the materials to a list might fall short pretty quickly

Lemmons:
@paxx12 Awesome. Thanks.  I'll see if I can get those changes made in the next day or two.

Lemmons:
@paxx12 I had to do a pretty major rework of it, but I believe I have incorporated the changes requested.  I'll post some new screenshots before too long of what the new ui looks like.

paxx12:
@Lemmons Awesome

Lemmons:
@paxx12 I updated the pr description with the updated screenshots.  There are still a few rough edges (such as the code currently not being able to tell the difference between an empty tag and an invalid/corrupted tag), but I think we can deal with those later.

paxx12:
@Lemmons we need to think how to connect those also with @suchmememanyskill changes. I think probably `gcode macro` should be enough. I will for sure make amendments. I will think about that.

suchmememanyskill:
Yeah, adding in the gcode should be enough. Although currently my filament detection system cannot write. Can see if i can implement that.

paxx12:
@Lemmons I'm thinking that the best would be to drop all special handling of OpenSpool from firmware, just expose raw command to write data to tag. So, make `html/js` to generate valid payload, but call: `FILAMENT_TAG_WRITE CHANNEL=N DATA=HEXENCODED`.

I have my own `ndef.js` encoder in here: https://github.com/paxx12/PrintTag-Web/blob/main/public/ndef.js.

Lemmons:
> @Lemmons I'm thinking that the best would be to drop all special handling of OpenSpool from firmware, just expose raw command to write data to tag. So, make `html/js` to generate valid payload, but call: `FILAMENT_TAG_WRITE CHANNEL=N DATA=HEXENCODED`.
> 
> I have my own `ndef.js` encoder in here: https://github.com/paxx12/PrintTag-Web/blob/main/public/ndef.js.

That sounds reasonable.  Would make it a bit more flexible to any tag type.

paxx12:
@Lemmons I'm thinking if also the parsing could not be done on `index.html` as well. I could expose raw data as well. But lets not change this yet. As if parsing is also done on web, then you could pretty much implement many formats, and work well across many RFID streams, even esp32 making this more universal than just U1.

Lemmons:
@paxx12 Do you want me to make the changes to the write tag gcode command and javascript to do the tag data generation on the js side?  If so I can work on that now

paxx12:
@Lemmons Yes, write side.

Lemmons:
there is a somewhat minor tag corruption bug right now, where if you write some arbitrary data to the tag, the delete and write logic might not succeed.  I'm working on this bug, but at least for now this has all the business logic needed.

overlays/firmware-extended/13-rfid-support/root/home/lava/klipper/klippy/extras/filament_protocol_ndef.py
```diff
@@ -234,15 +276,27 @@ def openspool_parse_payload(payload, card_uid=[]):
         return filament_protocol.FILAMENT_PROTO_ERR, None
 
 def ndef_proto_data_parse(data_buf):
-    error, records, card_uid = ndef_parse(data_buf)
+    error, records, card_uid, cc_bytes = ndef_parse(data_buf)
+
+    # Expected NTAG215 CC: E1 10 3F 00 (NDEF magic, v1.0, 504 bytes, read/write)
+    # Expected NTAG216 CC: E1 10 6D 00 (NDEF magic, v1.0, 872 bytes, read/write)
```
Lemmons: other parts of the code don't support ntag216, but it won't hurt anything here.

Lemmons:
@paxx12 I think this is read for review, if you don't have any other changes you'd like to see

Lemmons:
one item of potential discussion, I used urlsafe base64 encoding for the raw tag data payload in the gcode command, but I'd be find changing it to HEX, this is more of a "technically its more efficient during transmission" decisions which was unneeded.

overlays/firmware-extended/13-rfid-support/patches/04-add-card-type-tracking.patch
```diff
@@ -0,0 +1,26 @@
+  Including: home/lava/klipper/klippy/extras/filament_detect.py
+--- rootfs.original/home/lava/klipper/klippy/extras/filament_detect.py	2026-01-19 20:46:51.443456566 +0000
++++ rootfs/home/lava/klipper/klippy/extras/filament_detect.py	2026-02-08 22:00:00.000000000 +0000
```
paxx12, line 3: It seems it would be better to merge this change with prior patch, since this now creates patches that has to run in a specific sequence.

overlays/firmware-extended/13-rfid-support/root/home/lava/default-config/extended/klipper/05_filament_tag.cfg
```diff
@@ -0,0 +1,4 @@
+# NTAG tag write/erase operations
+# Provides gcode commands for writing and erasing NTAG filament tags
+
+[filament_tag]
```
paxx12, line 4: I would likely add this as a klipper tweak, feature that has to be explicitly enabled.

overlays/firmware-extended/13-rfid-support/root/home/lava/klipper/klippy/extras/filament_protocol_ndef.py
```diff
@@ -42,19 +56,24 @@ def ndef_parse(data_buf):
...
 
+        # Find the CC (Capability Container) page
+        # On NTAG tags read from page 0, CC is at byte 12 (page 3)
+        # CC magic is 0xE1, but OTP corruption may set extra bits (e.g. 0xF5)
+        # Check with bitmask: (byte & 0xE1) == 0xE1
         start_offset = 0
         if len(data) > 12 and data[0] != 0xE1:
             for i in range(min(16, len(data) - 4)):
-                if data[i] == 0xE1 and (data[i+1] == 0x10 or data[i+1] == 0x11 or data[i+1] == 0x40):
+                if (data[i] & 0xE1) == 0xE1 and (data[i+1] & 0x10) == 0x10:
```
paxx12, line 66: Seems that only this change is required in this file? Correct as this improved ndef parsing? Can we drop all rest of changes? Can we put this change in separate commit?

overlays/firmware-extended/13-rfid-support/root/home/lava/klipper/klippy/extras/filament_tag.py
```diff
@@ -0,0 +1,283 @@
...
+        while self.reactor.monotonic() < deadline:
+            if self.filament_detect._state[channel] == FILAMENT_DT_STATE_IDLE:
+                return True
+            self.reactor.pause(self.reactor.monotonic() + 0.1)
+        return False
+
+    def _ensure_cc_bytes(self, channel, info):
+        """Ensure CC bytes (page 3) are correct for an NTAG tag.
+
+        CC is OTP (One-Time Programmable): hardware performs new = old | written,
```
paxx12, line 64: Interesting. Forbid my ignorance, but can we erase the tag to write it next without performing this form of re-detect?

overlays/firmware-extended/13-rfid-support/root/home/lava/www/rfid-manager/index.html
```diff
@@ -0,0 +1,164 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>RFID Tag Manager - Snapmaker U1</title>
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@simonwep/pickr/dist/themes/nano.min.css">
```
paxx12, line 7: I would really want to avoid external dependencies. Is this required? Or can we generate a minimal CSS?

overlays/firmware-extended/13-rfid-support/root/home/lava/www/rfid-manager/index.html
```diff
@@ -0,0 +1,164 @@
+<!DOCTYPE html>
```
paxx12, line 1: @Lemmons I think we need to test this after enabling the Web Authentication. Can you take a look at `auth.js` use in `firmware-config` and `remote-screen` to read, inject bearer token and show when the authentication required?

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "deps/screen-apps"]
 	path = deps/screen-apps
 	url = https://github.com/paxx12/screen-apps.git
+[submodule "deps/PrintTag-Web"]
```
paxx12, line 4: I'm still debating with myself about `submodules` or `cache_git.sh`.

paxx12 (changes requested):
I have some initial questions. I did not yet run this myself to test exactly how it behaves. 

Random thought:

This is very interesting project, it just asks if it has to be part of this repo, or could it be managed externally.

I'm starting to think if this would not be better suited to be connected with `filament-detect` together and built to provide its own interface and actually provide the UI for that project.

This way both, would be treated as external dependency.

cc @suchmememanyskill

overlays/firmware-extended/13-rfid-support/patches/04-add-card-type-tracking.patch
Lemmons: yeah, I wasn't sure what the preference was for these sort of changes, but I'm fine with that.

overlays/firmware-extended/13-rfid-support/root/home/lava/klipper/klippy/extras/filament_tag.py
Lemmons: yeah.  This is mainly just to make sure that the CC bits are set to the expected values, and raise a warning if not.  My understanding is that they should generally be set to the correct values by the manufacturer, but I haven't had a chance to go through and look at it in too much detail.  When I was initially playing around with writing tags I corrupted the CC bits of a few of them, which lead me down a bit of a rabbit hole.  A bit of the CC between both extras files can likely be cleaned up a bit, but likely a decent amount of it can be cleaned up, or regulated to more generic warning logs rather than method outputs, etc.

Lemmons: definitely doesn't need to be done when erasing/rewriting an existing good tag.

overlays/firmware-extended/13-rfid-support/root/home/lava/www/rfid-manager/index.html
Lemmons: Sounds good.  Auth actually is one my specialties😅

overlays/firmware-extended/13-rfid-support/root/home/lava/www/rfid-manager/index.html
Lemmons: this is just the color picker.  The native color picker doesn't have transparency support, and I wanted to be able to support a color picker with transparency.  It would not be too difficult to just write a color picker with transparency support ourselves though.

.gitmodules
Lemmons: Personally, I really don't like submodules, but usually when I've used them it's in large, complex repos, where they wreak havoc.  Here I'm happy to do whatever.

Lemmons:
> I have some initial questions. I did not yet run this myself to test exactly how it behaves.
> 
> Random thought:
> 
> This is very interesting project, it just asks if it has to be part of this repo, or could it be managed externally.
> 
> I'm starting to think if this would not be better suited to be connected with `filament-detect` together and built to provide its own interface and actually provide the UI for that project.
> 
> This way both, would be treated as external dependency.
> 
> cc @suchmememanyskill

Yeah, the further along it has gotten, I'm starting to think the same.  That said, I don't know how much more time I can realistically put into it.  With that though, I want to see it in front of people and making the filament tagging ecosystem better, so for that reason alone I'd really like to see it to some form of completion.

@suchmememanyskill should I pivot this to support your [filament_detect](https://github.com/suchmememanyskill/filament-detect) project instead?  I could first port the tag writing logic over, and then we could put the UI as a front end for the project?

nuclearmistake:
I've been building with and running this on my local hackjob branch and it's fantastic.

Trying to write during a print will crash the printer currently. Needs this in 2 places in filament_tag.py to prevent arithmetic with None inside get_status
```         # Check if print is in progress
         print_stats = self.printer.lookup_object('print_stats', None)
-        if print_stats and print_stats.get_status(None).get('state') == 'printing':
+        if print_stats and print_stats.get_status(self.reactor.monotonic()).get('state') == 'printing':
```

convicte:
It would be fantastic to have this incorporated, since no matter how many different aproaches I've tested each is lacking across usability, spoolman integration, tag operability with U1, etc.

I don't see currently any pipeline that allows both filament quantity tracking, tag read and write as well as U1 tag reliable pickup.

Lemmons:
I've ran out of time to push this forward.  I'd be happy if anyone wanted to take it over and get it over the finish line.

paxx12:
I did rebase this on top of `develop`.

paxx12:
Superseded by #581 (Filament Manager UI).

This PR was the original inspiration and base for that work 0 thank you, @Lemmons. #581 deliberately drops writing/erasing tags from the printer: the design is to use Spoolman as the source of truth for spool data, with the printer assigning Spoolman spools to channels rather than programming physical RFID tags. Closing this in favour of that direction.

PR #158, opened 2026-01-19
Add SSH key embedding and dev.sh SSH mounting for development

Lemmons:
I don't think this is particularly in the correct place, but I wasn't sure what the best approach was here.  While working on the [rfid ui](https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/156) I wanted a way to be able to be a little more hands off when updating firmware using the update-firmware script.  This is the best I was able to come up with.  I assume there is a much better way to do this though (I imagine if we do want to keep the ability bake in ssh keys, it should go in the devel overlays, rather than common).  Definitely would like to get feedback on how we can make the iteration process faster when you need to push new versions to a real machine, and will be happy to update this as needed to match what that process might be.

# Changes
- Add overlays/common/04-ssh-authorized-keys/ overlay for embedding SSH public keys
- Add SSH folder mounting to dev.sh for container development
- Update .gitignore to exclude personal SSH keys
- Update upgrade-firmware.sh to remove the ssh host from known hosts, as it will fail due to the firmware upgrade changing the host sha

### DIFF
--- a/scripts/dev/upgrade-firmware.sh
+++ b/scripts/dev/upgrade-firmware.sh
@@ -11,6 +11,13 @@ shift 2
 
 set -xe
 
+# this will remove any old host fingerprints for the host being updated
+# and will quickly connect to it to allow the user to confirm the new fingerprint
+# this avoids any ssh prompts during the build and upgrade process (if the user has an ssh key setup)
+# this is inherently unsafe, so should only be used in controlled environments
+ssh-keygen -f /root/.ssh/known_hosts -R $SSH_HOST
+ssh $SSH_HOST exit
+
 rm -rf "firmware/firmware_$PROFILE.bin" "tmp/firmware"
 make build OUTPUT_FILE=firmware/firmware_$PROFILE.bin PROFILE="$PROFILE"
 scp "tmp/firmware/update.img" "$SSH_HOST:/tmp/"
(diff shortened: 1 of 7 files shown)


scripts/dev/upgrade-firmware.sh
```diff
@@ -11,7 +11,14 @@ shift 2
 
 set -xe
 
+# this will remove any old host fingerprints for the host being updated
+# and will quickly connect to it to allow the user to confirm the new fingerprint
+# this avoids any ssh prompts during the build and upgrade process (if the user has an ssh key setup)
+# this is inherently unsafe, so should only be used in controlled environments
+ssh-keygen -f /root/.ssh/known_hosts -R $SSH_HOST
```
mcristina422: I like what you're going for here, I end up doing the same thing manually. 
I think long term, we should reconfigure dropbear SSH server to save and reuse it's SSH Host Keys in a persistent mount point, so its not regenerated every upgrade

paxx12:
@Lemmons 

I get the idea, but until there are fixed credentials on printer this is also not needed. For example this is what I'm doing:

```bash
sshpass -p snapmaker ssh-copy-id root@u1eth.home && ./scripts/dev/upgrade-firmware.sh root@u1eth.home extended
```

paxx12:
Replaced by https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/222 there's also the https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/208.

PR #171, opened 2026-01-22
add tailscale binaries and config

mcristina422:
This installs the [tailscale](https://tailscale.com/) binaries and sets up a basic init script to run it in userspace mode. 

Tailscale must be first enabled by updating the tailscale enabled variable in `extended.cfg`. Then you can ssh to your printer and run `tailscale up` to complete the login flow. 

Later, to avoid any SSH, it would be possible to configure flags (like `--auth-key`) in the `extended.cfg`. I think that can be added in a followup PR. 

State is stored in `/home/lava/printer_data/tailscale/` to persist through reboots.


overlays/firmware-extended/22-tailscale/root/etc/init.d/S22-tailscale
```diff
@@ -0,0 +1,66 @@
...
+DAEMON="/usr/sbin/tailscaled"
+PIDFILE="/var/run/$NAME.pid"
+STATE_DIR="/home/lava/printer_data/tailscale"
+
+# Load config from user
+EXTENDED_CFG="/home/lava/printer_data/config/extended/extended.cfg"
+TAILSCALE_ENABLED=$(/usr/local/bin/extended-config.py "$EXTENDED_CFG" tailscale enabled false)
+
+# Skip when not enabled
+[ "$TAILSCALE_ENABLED" = "true" ] || exit 0
```
paxx12: Please move this to `start()`. We need to still shutdown the service if user disabled it.

paxx12 (changes requested):
Overall this looks amazing.

PR #220, opened 2026-01-31
Small change to update the fake-service make file to use the correct compiler based on the target arch

QuinnDamerell:
That's all.


paxx12:
Thanks!

PR #225, opened 2026-02-01
Klipper commits 16fc46f + 16b4b6b | Patch redone for U1 FW 1.1

Argolein:
I have redone the patches for FW 1.1. Instead of 5(?) patches I combined it into two patches. 
Tested the resonance compensation and printing as well. No errors or unwanted behaviour.

I do not work with patch files at all and find it still irritating , so I hope this is how it should be. 

Added origin as well.

Used config:

[resonance_tester]
accel_per_hz: 60
min_freq: 5
max_freq: 133.33
sweeping_accel: 400
sweeping_period: 1.2

Resonance compensation results:
calibration_data_x_20260201_103725
calibration_data_y_20260201_103725


paxx12:
@Argolein 

Thank you for this PR.

The purpose of 5 patches is to track exact changes from Klipper. The idea of redoing is interesting, but also this is not usually how backports are being done. The folded patches means that this is complete reimplementation of that feature based on the upstream, but there's no exact backing of those to commits in upstream. For example the `16b4b6b` depends on prior patches being present this is why they were added.

To better help me understand how this exactly changes things could you help me with those questions?

- Can you provide a diff of changes between current `develop` and your patches.
- Can you also provide me graphs from the current `develop` vs your changes?
- Can you also provide explanation for used values in `[resonance_tester]` are those required or can the defaults be used?

PR #236, opened 2026-02-04
Add Multi-tool support in Fluidd/Mainsail by emulating AFC

paxx12:
## Summary

Make U1 to behave as AFC enabled with Spoolman integration for multi-material printing support

### What it does

- **AFC status in UI** - Lane information (material, color, vendor, spool ID) visible in AFC-compatible frontends
- **Spoolman integration** - Assign spools to lanes with `SET_SPOOL_ID LANE=E0 SPOOL_ID=123`, automatically fetches filament data
- **Active spool tracking** - Current lane's spool ID synced to Spoolman
- **RFID spool detection** - OpenSpool tags with `spool_id` field link physical spools to Spoolman entries

### How it works

The AFC stub module is a read-only facade over the printer's existing `print_task_config` data structure. It does not implement its own filament tracking - all lane properties (material, color, vendor) are read from and written to the internal configuration that the printer already uses. Macros like `CHANGE_TOOL` and `LANE_UNLOAD` call the existing `AUTO_FEEDING` command.

Spoolman queries go through Moonraker's spoolman component via a new proxy RPC, with responses applied to `print_task_config` using the standard `SET_PRINT_FILAMENT_CONFIG` command.

### Configuration

Enable via firmware-config web interface under `Tweaks → AFC Stub`. Then edit `moonraker/spoolman.cfg` to enable a spoolman integration.

### Caveats

- The AFC component can only map lane (extruder) to a tool. There's no support to map many tools to a single lane (extruder) as Snapmaker Orca is doing.
- The runout logic cannot be controlled, as this is not configurable by Snapmaker, so it fails now.
- The weight tracking is not working if not used in Spoolman mode.
- The list of tools is derived fom `lane.map` instead of defined explicitly, so user can choose only from `T0..T3` tools.

### Integration with Snapmaker Orca

snorca_sync

### Re-order tools

edit_tools

### RFID support

rfid_read

### Spoolman support

spoolman

### DIFF
--- a/docs/rfid_support.md
+++ b/docs/rfid_support.md
@@ -62,7 +62,8 @@ Example payload:
   "min_temp": 190,
   "max_temp": 220,
   "bed_min_temp": 50,
-  "bed_max_temp": 60
+  "bed_max_temp": 60,
+  "spool_id": 1
 }
 ```
 
@@ -102,6 +103,7 @@ Using the non-standard OpenSpool `subtype` field it is possible to specify a mat
 - `additional_color_hexes` - Additional colors for multicolor spools (up to 4)
 - `weight` - Spool weight in grams
 - `diameter` - Filament diameter in mm (e.g., 1.75)
+- `spool_id` - Spoolman spool ID for integration with spool management systems
 
 ### Snapmaker Orca Naming Convention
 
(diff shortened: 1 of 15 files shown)


paxx12:
Replaced by https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/285

Issue #247, opened 2026-02-06
Proposal: Optional Klipper Router mode (multi-Klippy scaffolding)


justinh-rahb:
### Summary

I’d like to propose adding an **opt-in “Klipper Router mode”** to Extended Firmware that provides first-class scaffolding for running [**klipper-router**](https://github.com/paxx12/klipper-router) and **additional Klippy instances**, without assuming any specific peripherals (LED/MMU/etc.).

This has been **implemented and validated on a different platform (Ender 3 V3 SE)** and tested with:

* a **vendor-modified Klipper** running on the printer mainboard, and
* **stock, unmodified Klipper** running on a secondary RP2350 MCU.

The goal is to ship the **capability + safe lifecycle management**, while keeping klipper-router fully drop-in and requiring **no main-Klipper modifications**.

---

## High-level intent

When enabled:

* firmware manages `klipper-router` as a service
* firmware can start/stop **additional Klippy instances**
* instances can fail, restart, or disconnect independently
* router derives readiness by observing standard Klipper status objects
* primary motion Klipper remains unaffected
* everything is reversible via a single toggle

When disabled:

* system returns to standard single-Klippy behavior

No Klipper forks, no patches, no required macros or config loaded into the main instance.

---

## Validated behavior (current state)

* multiple Klippy instances run concurrently
* router detects instance disconnect/reconnect
* peripheral MCU firmware restart does **not** crash primary Klipper
* router operates without any custom signaling from main
* works with:

  * modified/vendor Klipper on mainboard
  * stock Klipper on secondary MCU
* no changes required to Klipper itself

---

### Extended Firmware (platform / UX)

Owns:

* feature flag + UI toggle (likely in `extended2.cfg`)
* service lifecycle (router + extra Klippy instances)
* filesystem conventions
* instance discovery / autostart
* observability (status + logs)
* safe rollback if router fails

Does **not** own:

* routing semantics
* readiness detection
* G-code APIs
* event logic

---

### klipper-router

Owns:

* routing behavior
* connect / disconnect / reconnect semantics
* readiness detection (derived from standard status objects like `print_stats`, `idle_timeout`, etc.)
* message & event bridging
* router-specific config parsing
* global router methods (`router/gcode_script`, `router/on_event`, etc.)

No assumptions about systemd, UI, or firmware layout.

---

### User land

Owns:

* defining extra Klippy instances
* assigning MCUs to instances
* router topology/mappings
[...]

PR #255, opened 2026-02-08
firmware-extended: add opt-in klipper-router mode with isolated instance orchestration

justinh-rahb:
**Description**  
This PR builds on #247 and adds an opt-in Klipper Router mode to Extended Firmware to separate failure domains between core motion control and optional peripherals.

Today, optional subsystems (MMU/AMS/feeder/sensors/toolchanger controllers) often share the same Klippy failure domain as motion control. A peripheral reset/timeout can abort an otherwise healthy print. This change introduces infrastructure for running peripherals as independent Klippy instances, with routing between instances handled by [`klipper-router`](https://github.com/paxx12/klipper-router).

No Klipper fork or core motion-control patching is required.

**What this PR adds**

- New overlay: `overlays/firmware-extended/25-u1-router-led-events`
- Pinned install of upstream `klipper-router` during firmware build
  - daemon: `/usr/local/sbin/klipper-routerd`
- Opt-in router mode toggle in firmware-config
  - `functions/25_settings_router.yaml`
  - setting: `[router] enabled` in `extended2.cfg`
  - basic runtime status visibility (router service, instance count, log path)
- Service orchestration
  - `S98klipper-router-instances`: discovers and runs additional Klippy instances from:
    - `/home/lava/printer_data/config/extended/router/instances/<name>/printer.cfg`
    - per-instance sockets: `/home/lava/printer_data/comms/klippy-router-<name>.sock`
  - `S99klipper-router`: runs router daemon and builds runtime router config with discovered instances
- Default router filesystem layout under extended config defaults
  - router config: `extended/router/klipper_router.cfg`
  - default LED instance template and macros:
    - `extended/router/instances/led/**`
- Extended-side router macro API + LED status event consumption
  - `extended/klipper/15_router_api.cfg`
  - `extended/klipper/16_router_led_event_subscriptions.cfg`
- Boot-time migration hook for persisted config
  - `S50router-led-events` ensures LED re-subscribe hook exists in `15_router_api.cfg`
- LED instance MCU path updated to stable USB `by-path` device on U1 template

**Why this helps**

- Peripheral faults can be isolated from the main motion-control Klippy process.
- Non-critical subsystems can restart/reconnect without forcing a full print abort by default.
- Enables scaling to multiple optional controllers without tightly coupling everything into one Klippy instance.

**Scope boundaries**

This PR provides infrastructure only:
- router mode enablement
- service orchestration
- instance discovery/layout
- status visibility
- one concrete LED routed-state example

It does **not** define MMU/device policy semantics or rewrite print lifecycle logic globally.

**Operational model**

- Default state: disabled (`[router] enabled: false`)
- Enabling router mode starts isolated instance service + router service
- Disabling router mode stops both cleanly

**Validation focus for review**

- Enable/disable path via firmware-config
- Service start order and idempotency (`S49` -> `S50` -> `S98` -> `S99`)
- Router and instance socket/log creation
- LED routed status event state updates and duplicate suppression
- Clean behavior when auxiliary instance disconnects/reconnects

### DIFF
--- a/overlays/firmware-extended/25-u1-router-led-events/README.md
+++ b/overlays/firmware-extended/25-u1-router-led-events/README.md
@@ -1,0 +1,52 @@
+# 25-u1-router-led-events
+
+Installs Klipper Router and adds LED status-event consumption for U1 extended Klipper config.
+
+## What it does
+
+- Installs pinned `klipper-router` from `https://github.com/paxx12/klipper-router` during firmware build:
+  - daemon: `/usr/local/sbin/klipper-routerd`
+  - reference include: `/usr/local/share/firmware-config/router/includes/router_api.cfg`
+- Installs init service:
+  - `/etc/init.d/S98klipper-router-instances` (starts/stops additional Klippy instances)
+  - `/etc/init.d/S99klipper-router`
+- Adds default router runtime config:
+  - `/usr/local/share/firmware-config/extended/router/klipper_router.cfg`
+  - copied to `/home/lava/printer_data/config/extended/router/klipper_router.cfg` on boot
+- Adds default additional-instance layout:
+  - `/usr/local/share/firmware-config/extended/router/instances/led/printer.cfg`
+  - `/usr/local/share/firmware-config/extended/router/instances/led/klipper/*.cfg`
+  - copied to `/home/lava/printer_data/config/extended/router/instances/led/` on boot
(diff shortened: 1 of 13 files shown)


justinh-rahb:
Cleaning up old branches, this could be revisited in a new PR if anything new develops.

PR #272, opened 2026-02-13
Add AFC Lite as thin layer to simulate `ArmoredTurtle AFC-Klipper-Add-On`

paxx12:
Cut from the https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/236, without Spoolman support.

Implements minimal `AFC` Python modules that expose AFC-compatible status to enable Fluidd/Mainsail AFC UI panels. This is a stub/thin layer that simulates https://github.com/ArmoredTurtle/AFC-Klipper-Add-On/ without requiring actual AFC hardware.

Maps 4 lanes to U1 extruders with filament info from `print_task_config`. Includes macros (`CHANGE_TOOL`, `LANE_UNLOAD`, `SET_COLOR`, `SET_MATERIAL`) that wrap U1's `AUTO_FEEDING` and filament configuration commands.

### Integration with Snapmaker Orca

snorca_sync

### Re-order tools

edit_tools


2wDragon:
Is the support for Spoolman integration with the AFC is planed for the next relesed? 
Is it planed to have sync from the spoolman to the printer extroder like a virtual RFID so when I select spool in the spoolman that it sync to the extroder just like the RFID Tag do?

PR #276, opened 2026-02-14
Enable OctoEverywhere companion mode

paxx12:
Snapmaker U1 does not provide standard access like full-system Klipper installs, requiring companion mode. Remove MoonrakerConfigFile setting as companion mode does not modify moonraker.conf.


paxx12:
cc @QuinnDamerell

PR #285, opened 2026-02-15
Spoolman Idea 1: Add OpenSpool `spool_id` extension to enable Spoolman in AFC-Lite

paxx12:
This PR adds Spoolman integration to AFC and extends the RFID filament protocol to support `spool_id`.

**Workflow:** This requires a specially programmed NFC tags to include `spool_id`.

## Changes

- **RFID Protocol**: Added `spool_id` field to filament protocol NDEF for tracking spools
- **AFC Integration**: Enhanced AFC lane management with Spoolman API support for automatic spool tracking
- **Moonraker Proxy**: Added Spoolman proxy configuration to Moonraker
- **Configuration**: New AFC config with Spoolman-related macros and settings

### RFID support

rfid_read

### Spoolman support

spoolman


matissrapsa:
Hey @paxx12 any chanche this will be included in v1.2.0-paxx12-14

paxx12:
Unlikely, as this requires updates in various places, but after this, likely yes.

paxx12:
# Why `spool_id` (PR #285) was closed in favour of `card_uid` (PR #364)

## Problem with the `spool_id` approach

PR #285 extends the RFID NDEF record with a `spool_id` field that maps directly
to a Spoolman spool. The mapping lives *on the tag*, which creates several issues:

- The tag must be **specially programmed** with a `spool_id` - only works with
  OpenSpool-compatible tooling (SpoolPainter, etc.).
- **Any tag not programmed this way is excluded** - blank tags, manufacturer
  tags, read-only or undecryptable tags cannot participate.
- The tag is the source of truth, not Spoolman. This makes migration,
  re-binding, and tag replacement unnecessarily painful.
- Maintaining two parallel identification paths adds complexity and test surface
  with limited benefit.

## Why `card_uid` in Spoolman (PR #364) is correct

PR #364 flips the model: instead of storing the Spoolman spool ID on the tag,
it stores the tag's hardware `CARD_UID` in Spoolman's `lot_nr` field. The field is subject to be defined.

- **Spoolman becomes the single source of truth** - not the tag.
- **Any NFC tag works** - blank, pre-programmed, OpenSpool, read-only - as long
  as the reader can obtain its UID. The tag does not need to be writable,
  decryptable, or compliant with any particular NDEF schema.
- **First-time setup**: assign a spool to a lane via Fluidd once, firmware
  binds the `CARD_UID` into `lot_nr`. Every subsequent insertion is automatic.
- **Future-proof**: the same `lot_nr` lookup can extend to barcodes, QR codes,
  serial numbers, or any unique identifier - no firmware changes needed.
- **Mobile apps**: can update Spoolman database already, since they already fetch available spools from Spoolman.
- **Mobile apps are not required**: user might or might not use Mobile apps, since tag is never written by printer, the printer can create the mapping as well.

## End-user experience

Both approaches produce the same outcome at print time: insert a spool, the
printer resolves it to a Spoolman entry and tracks usage automatically.
The difference is invisible during printing. The only practical change is the
one-time setup - a UI click instead of physical tag programming.

## Migration from `spool_id` tags

Spools programmed with OpenSpool `spool_id` data continue to be read for
color/material metadata. The `card_uid` mapping in #364 uses the tag UID as the
key instead of `spool_id`. Re-binding is done through the normal UI flow
(one click per spool side), or via a planned mobile migration tool that reads
the UID and updates Spoolman's `lot_nr` in bulk.

## Long story short

This follows the same pattern as explained in here: https://help.simplyprint.io/en/article/all-about-nfc-rfid-for-filament-spools-bambu-openprinttag-creality-qidi-anycubic-more-19luyni/#3-the-problem-that-this-solves

PR #287, opened 2026-02-16
Improve RFID tag discovery reliability

paxx12:
Increase timeout from 10ms to 50ms in REQA/WUPA commands to allow more time for NTAG215 tags to respond.
Reduce retry attempts from 10/15 to 5/5 to optimize discovery speed.


paxx12:
This was confirmed by user as not needed.

PR #303, opened 2026-02-20
Expose RFID via `filament_detect/set`

paxx12:
- Document internal client contract and state polling.
- Allow to disable `fm175xx_reader`.

### DIFF
--- a/overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
+++ b/overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
@@ -1,0 +1,72 @@
+--- rootfs.original/home/lava/klipper/klippy/extras/filament_detect.py
++++ rootfs/home/lava/klipper/klippy/extras/filament_detect.py
+@@ -52,6 +52,9 @@
+         self.printer.register_event_handler("filament_feed:port", self._feed_port_evt_handle)
+         self.printer.register_event_handler("filament_switch_sensor:runout", self._runout_evt_handle)
+ 
++        self.webhooks = self.printer.lookup_object('webhooks')
++        self.webhooks.register_endpoint("filament_detect/set", self._handle_filament_detect_set)
++
+     def _ready(self):
+         self.filament_feed_objects = self.printer.lookup_objects('filament_feed')
+         self._fm175xx_reader = self.printer.lookup_object('fm175xx_reader')
+@@ -173,6 +176,59 @@
+ 
+         return error, info
+ 
++    def _handle_filament_detect_set(self, web_request):
++        try:
++            channel = web_request.get_int('channel', None)
++            if channel is None:
++                raise web_request.error("channel must be specified!")
++
++            if (channel < 0 or channel >= self._channel_nums):
++                raise web_request.error("channel[%d] is out of range[0, %d]"
++                                        % (channel, self._channel_nums - 1))
++
++            params = web_request.get_dict('info', {})
++            has_params = len(params) > 0
++            filament_info = copy.deepcopy(filament_protocol.FILAMENT_INFO_STRUCT)
++
++            if 'VENDOR' in params:
++                filament_info['VENDOR'] = str(params.pop('VENDOR'))
++            if 'MAIN_TYPE' in params:
++                filament_info['MAIN_TYPE'] = str(params.pop('MAIN_TYPE'))
++            if 'SUB_TYPE' in params:
++                filament_info['SUB_TYPE'] = str(params.pop('SUB_TYPE'))
++
++            if 'HOTEND_MIN_TEMP' in params:
++                filament_info['HOTEND_MIN_TEMP'] = int(params.pop('HOTEND_MIN_TEMP'))
++            if 'HOTEND_MAX_TEMP' in params:
++                filament_info['HOTEND_MAX_TEMP'] = int(params.pop('HOTEND_MAX_TEMP'))
++            if 'BED_TEMP' in params:
++                filament_info['BED_TEMP'] = int(params.pop('BED_TEMP'))
++
++            if 'ALPHA' in params:
++                filament_info['ALPHA'] = int(params.pop('ALPHA')) & 0xFF
++            if 'RGB_1' in params:
++                filament_info['RGB_1'] = int(params.pop('RGB_1')) & 0xFFFFFF
++            filament_info['ARGB_COLOR'] = filament_info['ALPHA'] << 24 | filament_info['RGB_1']
++
++            if 'CARD_UID' in params:
++                filament_info['CARD_UID'] = [int(b) for b in params.pop('CARD_UID')]
++            if 'SKU' in params:
++                filament_info['SKU'] = int(params.pop('SKU'))
++
++            if params:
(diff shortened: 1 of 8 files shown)


paxx12:
Replaces https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/256

kwas-oc:
aren't you missing passing other data? 

MANUFACTURER, SUB_TYPE, COLOR_NUMS, WEIGHT + handling of min/max of bed temp

paxx12:
@kwas-oc I will at least now expose minimal amount of fields that are actually supported by SnOrca. More can be added later. We don't know if Snapmaker will not change those before they start using them.

kwas-oc:
ok, I thought that we'll align on the existing code in FW as many more fields are processed in vanilla firmware. but also true, that we can add more if we find that they are used anywhere

paxx12:
They are not. They are live interpreted by SnOrca, but this is rather loose mapping, and not very stable.

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
```diff
@@ -0,0 +1,72 @@
...
++                color = str(params.pop('color')).strip()
++                filament_info['RGB_1'] = int(color, 16) & 0xFFFFFF
++            filament_info['ARGB_COLOR'] = filament_info['ALPHA'] << 24 | filament_info['RGB_1']
++
++            if 'card_uid' in params:
++                filament_info['CARD_UID'] = list(bytes.fromhex(str(params.pop('card_uid'))))
++            if 'sku' in params:
++                filament_info['sku'] = int(params.pop('sku'))
++            elif len(params) > 0:
++                filament_info['sku'] = rand()
```
wasikuss: why? if the params are not empty, then it will still raise an exception in line 60

```suggestion
+            filament_info['sku'] = int(params.pop('sku', rand()))
```

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
```diff
@@ -0,0 +1,72 @@
...
++
++            if params:
++                raise web_request.error("unsupported fields: %s" % (', '.join(sorted(params.keys()))))
++
++            info['OFFICIAL'] = len(params) > 0
```
wasikuss: as comment above, if params is not empty, it will raise an exception in line 60, so won't reach this line

paxx12: Correct. Will fix it in a next commit.

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
paxx12: This is intentionally kept empty for case when there's no params send.

wasikuss: I'm trying to point out that we have 3 cases now
1) params have sku -> it is set
2) params missing sku
   a) there is other params -> it is set but next will be raised exception
   b) there is no other params -> it remains empty

the problem with 2a - will never success. sku will never be assigned to random.

wasikuss: so whole if+elif can be shortened to `int(params.pop('sku', DEFAULT))` put DEFAULT to 0 if it should be empty in case of missing sku in params

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
```diff
@@ -0,0 +1,74 @@
...
++
++            if 'card_uid' in params:
++                filament_info['CARD_UID'] = list(bytes.fromhex(str(params.pop('card_uid'))))
++            if 'sku' in params:
++                filament_info['SKU'] = int(params.pop('sku'))
++
++            if params:
++                raise web_request.error("unsupported fields: %s" % (', '.join(sorted(params.keys()))))
++
++            filament_info['OFFICIAL'] = has_params
```
wasikuss, line 59: why we base this on amount of params?

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
```diff
@@ -0,0 +1,74 @@
...
++            self._state[channel] = FILAMENT_DT_STATE_IDLE
++
++            ptc = self.printer.lookup_object('print_task_config', None)
++            if has_params or ptc is not None and ptc.print_task_config['filament_official'][channel]:
++                self._filament_info_update(channel, filament_info, True)
++            web_request.send({'state': 'success'})
```
wasikuss, line 65: it returns success even if the previous condition can be false

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
paxx12: I will remove `sku` randomization, and only assign it when passed.

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
paxx12: this is to detect `clear` and `set` operations, the `OFFICIAL` has special internal behavior

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
paxx12: You mean this one?

```
+            if has_params or ptc is not None and ptc.print_task_config['filament_official'][channel]:
+                self._filament_info_update(channel, filament_info, True)
```

wasikuss: yes

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
wasikuss: I didn't checked this. So pushing empty info will do anything?

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
paxx12: This is valid, command still succeeded, but made no changes.

overlays/firmware-extended/13-rfid-support/patches/05-add-filament-detect-set-endpoint.patch
paxx12: Sending empty will clear RFID payload, and if RFID was configured, it will clear RFID data. But will not clear user-data when RFID clear is sent.

PR #313, opened 2026-03-02
Draft: Implement naive heated chamber via bed heating

wasikuss:
As in the title. Using bed heater to heat chamber. I'm thinking of adding also:
- [ ] hard limit to 70C? 80C?
- [ ] timeout to prevent infinite heating
- [ ] update docs how to enable it in slicer
- [ ] handle cancelling print from UI or screen
- [ ] *OPTIONALLY* allow to change `temperature_sensor` to something other than `cavity`

### DIFF
--- a/docs/tweaks.md
+++ b/docs/tweaks.md
@@ -86,6 +86,23 @@ Enables object processing in Moonraker's file manager to support adaptive mesh f
 **Configuration:**
 This feature can **only** be configured via Firmware Configuration web interface. Manual configuration is not supported.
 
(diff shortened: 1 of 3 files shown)


paxx12:
@wasikuss Should we merge this?

wasikuss:
See the todo list. still not solved. The most crucial is this: `handle cancelling print from UI or screen` - cannot be done due to blocking while waiting for heating, so only emergency stop works. the rest I'll be updated today

PR #315, opened 2026-03-03
fix: resolve web auth 401 and rsync OpenSSL 1.1 incompatibility

### LINKED ISSUE #154: Missing library for rsync
Rsync is missing a library (running U1_extended_1.0.0-paxx12-11_upgrade.bin)

root@lava:/home/lava# rsync
rsync: error while loading shared libraries: libcrypto.so.1.1: cannot open shared object file: No such file or directory

root@lava:/home/lava# ldd /usr/local/bin/rsync
        linux-vdso.so.1 (0x0000007f9f1c0000)
        libcrypto.so.1.1 => not found
        libc.so.6 => /lib/libc.so.6 (0x0000007f9f022000)
        /lib/ld-linux-aarch64.so.1 (0x0000007f9f1c1000)

It would be useful to have a working rsync as I backup all my machines with this to my NAS.

### LINKED ISSUE #296: Web Authentication (Experimental) - HTTP 401: Unauthorized
Hi!

Enabling the "Web Authentication (Experimental)" will throw following error. But it will essentially protect the UI with logon page.

```
=== Updating Web Authentication (Experimental) ===
Setting: Enabled
Time: 2026-02-18 20:22:35
========================================

Applying changes...
Deleting existing admin user...
Enabling force_logins in Moonraker...
Restarting Moonraker...
Stopping moonraker...
OK
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
wait 0.2s
Starting moonraker...
Waiting for Moonraker to restart...
Creating new admin user...
Warning: Could not create admin user
{"error": {"code": 401, "message": "Unauthorized", "traceback": "Traceback (most recent call last):\n\n  File \"/usr/lib/python3.11/site-packages/tornado/web.py\", line 1692, in _execute\n\n  File \"/home/lava/moonraker/moonraker/components/application.py\", line 503, in prepare\n    self.current_user = await auth.authenticate_request(\n                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n\n  File \"/home/lava/moonraker/moonraker/components/authorization.py\", line 886, in authenticate_request\n    raise HTTPError(401, \"Unauthorized, Force Logins Enabled\")\n\ntornado.web.HTTPError: HTTP 401: Unauthorized (Unauthorized, Force Logins Enabled)\n"}}

========================================
SUCCESS: Setting updated successfully
```

The issue might be the fact that I first created new user under Settings->Authentication and then enabled the WebAuth. The login then accepts the user i have created there. Seems like this "sequence" is not expected.

Firmware used: v1.1.1-paxx12-13

kbaker827:
### Fix #296 – Web Auth 401 on user creation

**Root cause:** `force_logins` was being enabled and Moonraker restarted *before* the admin user was created. Once Moonraker restarts with `force_logins: true` and no users exist, it rejects every API call — including `POST /access/user` — with 401 Unauthorized.

**Changes** (`15_settings_security.yaml`):
- Create the admin user **first**, while Moonraker still accepts unauthenticated requests
- Only after successful user creation: write `authorization.cfg` and restart Moonraker to enforce logins
- Removed the now-unnecessary `sleep 10`
- Added `exit 1` on user-creation failure so the error is clearly reported

### Fix #154 – rsync missing `libcrypto.so.1.1`

**Root cause:** The rsync binary from Samba's CentOS 8 aarch64 repo was compiled against OpenSSL 1.1, which is absent on the Snapmaker U1 (OpenSSL 3 only).

**Changes** (`rsync_install.sh` + `Dockerfile`):
- Replace the pre-built CentOS 8 binary with a **cross-compiled build** from the Debian upstream rsync 3.4.1 source tarball (SHA256-verified)
- Links against OpenSSL 3 (`libssl-dev:arm64`), system zlib (`zlib1g-dev:arm64`), and system libpopt (`libpopt-dev:arm64`)
- Stubs `zlib/dummy.in` and `popt/dummy.in` to satisfy `config.status` (the Debian `+ds1` tarball strips both bundled libraries)
- Bumps rsync 3.2.7 → 3.4.1

---

### Pin pip3 package versions for reproducible builds
`apprise`, `wled`, and `PyYAML` were installed with no version constraint. Pinned to current latest: `PyYAML==6.0.3`, `apprise==1.9.7`, `wled==0.21.0`.

### New System Actions
- **Shutdown System** — safely powers off the printer via `/sbin/poweroff`; previously only Reboot existed

### New Service Actions
- **Restart Nginx** — reloads nginx config without dropping connections; useful after switching frontend
- **Restart Firmware Config** — restarts the `S99firmware-config` web UI process

### Storage Status
Added a **Storage** section to the status page showing free space on `/userdata` (logs/upgrades) and `/oem` (persistent config).

---

### Check for Updates (`Troubleshooting` section)
One-click update check that:
- Shows the currently installed build version and profile
- Fetches the GitHub releases API for the latest release name and tag
- Finds the matching `.bin` download URL for the current profile
- Prints the URL with copy-paste instructions for the **Firmware Upgrade (URL)** field

Previously users had to navigate GitHub manually, find the release, identify the right `.bin` for their profile, and copy the URL. Now it's a single button press.

### Tailscale Login (`Troubleshooting` section, VPN overlay)
Runs `tailscale up` and streams the output live in the UI. If the node key has expired or the device was never authenticated, Tailscale prints an auth URL that the user can open on any browser. If already connected, it exits immediately.

This solves a circular problem: Tailscale was often enabled *to avoid needing SSH*, but an expired session required SSH to re-authenticate.

### DIFF
--- a/.github/dev/Dockerfile
+++ b/.github/dev/Dockerfile
@@ -36,6 +36,7 @@ RUN dpkg --add-architecture arm64 && \
     apt-get install -y \
       libssl-dev:arm64 \
       zlib1g-dev:arm64 \
+      libpopt-dev:arm64 \
     && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /workspace
--- a/.github/dev/Dockerfile
+++ b/.github/dev/Dockerfile
@@ -35,6 +35,7 @@ RUN dpkg --add-architecture arm64 && \
     apt-get update -y && \
     apt-get install -y \
       libssl-dev:arm64 \
+      zlib1g-dev:arm64 \
     && rm -rf /var/lib/apt/lists/*
 
 WORKDIR /workspace
--- a/overlays/firmware-extended/01-system-utils/scripts/rsync_install.sh
+++ b/overlays/firmware-extended/01-system-utils/scripts/rsync_install.sh
@@ -32,13 +32,19 @@ tar -xJf "$CACHE_DIR/$FILENAME" -C "$BUILDDIR"
 
 SRC=$(ls -d "$BUILDDIR"/rsync-*)
 
+# The Debian +ds1 tarball strips the bundled zlib/ directory to comply with
+# Debian policy, but rsync's autoconf-generated config.status still references
+# zlib/dummy.in as a template even when --with-included-zlib=no is set.
+# Create a stub to satisfy config.status without affecting the actual build.
+mkdir -p "$SRC/zlib"
+touch "$SRC/zlib/dummy.in"
+
 echo ">> Cross-compiling rsync for aarch64..."
 (
   cd "$SRC"
   # rsync 3.4.1 uses --disable-* flags (not --without-*) for optional features.
   # OpenSSL is auto-detected via -lcrypto; no --with-openssl flag needed.
-  # The Debian +ds1 tarball strips the bundled zlib/, so we must use the system
-  # zlib (zlib1g-dev:arm64) via --with-included-zlib=no.
+  # --with-included-zlib=no forces use of system zlib (zlib1g-dev:arm64).
   ./configure \
     --host=aarch64-linux-gnu \
     CC=aarch64-linux-gnu-gcc \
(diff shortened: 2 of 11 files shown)


paxx12:
Sorry, this project expect the contributions to be primarily made by humans. AI can help making those, but this one does not indicate that was done by human and was additionally tested by human.

PR #325, opened 2026-03-06
feat(tweaks): add Cancel Individual Objects toggle

kbaker827:
## Summary

Adds a **Cancel Individual Objects** toggle to the Tweaks section of the firmware-config web interface. When enabled, Klipper's built-in `[exclude_object]` module is activated, allowing users to cancel individual objects mid-print from Fluidd or Mainsail without cancelling the entire print job.

- Follows the exact same pattern as the existing TMC tweaks (enable/disable writes/removes a cfg file and restarts Klipper)
- Zero overhead: no Moonraker-side object processing, no file upload delay
- Prints sliced without object labels work normally — the cancel UI simply doesn't appear
- Full docs added to `tweaks.md` including a slicer compatibility table

## How it works

Toggling to **Enabled** copies `exclude_object.cfg` (containing just `[exclude_object]`) into `/oem/printer_data/config/extended/klipper/` and restarts Klipper. Toggling to **Disabled** removes it.

## Slicer setup (user responsibility)

| Slicer | Where | Option |
|---|---|---|
| OrcaSlicer / Bambu Studio | Print Settings → Others | **Label objects** |
| PrusaSlicer | Print Settings → Output options | **Label objects** |
| Cura | Marketplace | **Exclude Objects** plugin |

## Test plan

- [ ] Enable the tweak; confirm `[exclude_object]` appears in `GET /printer/objects/list` response
- [ ] Slice a multi-object plate with "Label objects" enabled; confirm cancel buttons appear in Fluidd/Mainsail during print
- [ ] Slice without "Label objects"; confirm print works normally with no errors
- [ ] Disable the tweak; confirm `exclude_object.cfg` is removed and Klipper restarts cleanly

### DIFF
--- a/docs/firmware_config.md
+++ b/docs/firmware_config.md
@@ -50,7 +50,7 @@ Toggle settings directly from the web interface:
 | Klipper Metrics Exporter | Enabled, Disabled | Enable Prometheus metrics |
 | VPN Provider | None, Tailscale | Enable VPN remote access (Experimental) |
 | Cloud | None, OctoEverywhere | Enable Cloud-based remote access (Experimental) |
-| Tweaks | TMC AutoTune, TMC Reduced Current, Object Processing, AFC Stub | Experimental Klipper tweaks ([tweaks](tweaks.md)) |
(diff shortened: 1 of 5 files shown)


paxx12:
The `exclude_object` is intentionally not enabled, since it is not working properly. It skips colors on prime tower. Once it gets enabled in upstream firmware, this firmware will have it enabled as well.

PR #328, opened 2026-03-08
Fixes #209:  Add user-controlled toggle to suppress mqtt_agent DNS polling

### LINKED ISSUE #209: Remote screen not working on pr206 build
I have been using the 1.0.0 release for a few weeks and the remote screen was configured and working. Today, I installed the pr206 build and the remote screen no longer works.

Checked Advanced Mode (toggled off and back on several times with power cycle between)

Verified that extended.cfg and 04_remote_screen.cfg both have enabled=true as expected

fluidd shows No Signal / Maybe not enabled?

Case camera still works, as does whatever else I have tried in fluidd

### LINKED ISSUE #290: Two DNS requests every 5s
Hi,

unsure if this was present on any other firmware before or whether its a problem of your (very much appreciated extended firmware) or the original firmware: Currently, I am seeing two DNS requests every ~5s for id.snapmaker.com. One request for the A record, second request for the AAAA record.
Anybody else with the same observation?

All the best

IamMrCupp:
## Summary

Fixes #290 — the printer emits two DNS queries (A + AAAA) for `id.snapmaker.com` every ~5 seconds due to `mqtt_agent` cycling through its reconnect state machine when cloud auth fails.

- Adds a flag-file guard to `S59mqtt_agent` via a new common overlay (`05-disable-cloud-polling`), so the service can be suppressed without patching the closed-source binary. Applies to both `basic` and `extended` profiles.
- Adds a **Snapmaker Cloud Agent** toggle in the firmware-config UI under **Remote Access** (extended profile only), so users can enable/disable the service without SSH.
- Defaults to **enabled** — no breaking change for users who use the Snapmaker mobile app or cloud remote control.

## Root cause

`mqtt_agent` enters a `NETWORK_DETECTION` → `CONNECTING_BROKER` → failure → retry loop every ~5 seconds when the device is not enrolled in the Snapmaker cloud (or auth has expired). Each retry cycle calls `getaddrinfo("id.snapmaker.com")`, producing the A + AAAA query pattern. The binary accepts no flag to disable cloud connectivity.

## Changes

- **`overlays/common/05-disable-cloud-polling/patches/01-guard-mqtt-agent.patch`** — patches `S59mqtt_agent` to exit early if `/oem/.cloud_agent_disabled` exists. `/oem/` is persistent storage, so the setting survives reboots and firmware updates.
- **`overlays/firmware-extended/10-firmware-config/root/usr/local/share/firmware-config/functions/17_settings_cloud_agent.yaml`** — adds the UI toggle. Selecting **Disabled** writes the flag file and stops the running service live; **Enabled** removes the flag and starts it. No reboot required.

## Test plan

- [x] Flashed firmware to test printer, navigated to **Remote Access** in firmware-config — **Snapmaker Cloud Agent** toggle present, defaulting to **Enabled**
- [x] Confirmed baseline DNS queries to `id.snapmaker.com` (A + AAAA) visible every ~5 seconds while enabled
- [x] Selected **Disabled**, confirmed dialog — `mqtt_agent` stopped, `/oem/.cloud_agent_disabled` created, DNS queries ceased
- [x] Selected **Enabled** — service restarted, DNS queries resumed within ~10 seconds
- [x] Rebooted with flag present — `mqtt_agent` did not start on boot, flag persisted in `/oem/`
- [x] Basic profile SSH test: `touch /oem/.cloud_agent_disabled && /etc/init.d/S59mqtt_agent stop` — process stopped cleanly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,18 @@ tools/rk2918_tools/img_maker
 tools/rk2918_tools/img_unpack
 tools/rk2918_tools/mkkrnlimg
 _site/
+
+
+### macOS ###
+# General
+.DS_Store
+.AppleDouble
+.LSOverride
+
+# Thumbnails
+._*
+
+# Claude.ai files and stuff
+.claude
+plan.md
+research.md
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,3 @@ _site/
 
 # Claude.ai files and stuff
 .claude
-plan.md
-research.md
--- a/overlays/common/05-disable-cloud-polling/patches/01-guard-mqtt-agent.patch
+++ b/overlays/common/05-disable-cloud-polling/patches/01-guard-mqtt-agent.patch
@@ -1,0 +1,12 @@
+--- a/etc/init.d/S59mqtt_agent
++++ b/etc/init.d/S59mqtt_agent
+@@ -6,6 +6,9 @@
+ PROG=mqtt_agent
+ PIDFILE=/var/run/mqtt_agent.pid
+ ARGS="-v true"
++
++# Skip cloud agent if disabled by custom firmware
++[ -f /oem/.cloud_agent_disabled ] && exit 0
+
+ log()
+ {
--- a/overlays/firmware-extended/10-firmware-config/root/usr/local/share/firmware-config/functions/17_settings_cloud_agent.yaml
+++ b/overlays/firmware-extended/10-firmware-config/root/usr/local/share/firmware-config/functions/17_settings_cloud_agent.yaml
@@ -1,0 +1,32 @@
+settings:
+  remote_access:
+    items:
+      cloud-agent:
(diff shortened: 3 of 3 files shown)


paxx12:
Nice. You are thinking very similar way :) 

I have those changes in another branch to run `fake-mqtt-agent`, since GUI depends on some MQTT status: https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/124/changes#diff-ce6c35e4dc3afb85672f9c1acd204086f2ea90561a70c4eb2ebcb2112b2c3b17.


Maybe we should include this and also disable `system_mgmt`?

paxx12:
@IamMrCupp The SM did enable DNS caching in latest v1.2.0 firmware. Do we still need it in this way?

paxx12:
This change is no longer relevant, as latest firmware uses DNS caching.

PR #349, opened 2026-03-21
feat: add dev script for amd64 hosts

midnya:
I ran into issues when trying to build my own firmware, and edited `./dev.sh` to fix them. I was able to build a entire firmware, from nothing up to its deployment on my U1.

I can guarantee that this isn't *enough* to run on Linux amd64; for example,  I had to install `qemu-user-static` to run the image. As my system is already set up for cross-platform Docker image builds, there's likely even more steps missing.

Opened this PR to open the discussion!

### DIFF
--- a/dev.amd64.sh
+++ b/dev.amd64.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -e
+
+IMAGE_NAME="snapmaker-u1-dev"
+BUILD_CONTEXT=".github/dev"
+
+# See https://docs.docker.com/build/building/multi-platform/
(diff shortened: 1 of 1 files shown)


paxx12:
@ShinySaana Basically the cross-compile on `amd64` should work to avoid long build times when using `linux/arm64`. However, it does require `qemu-user-static` + `binfmt_misc` support loaded, to ensure that arm64 can be run on host platform.

midnya:
@paxx12 Retried; indeed it worked.
Should probably be added it in the docs somewhere? It's your call.

paxx12:
@ShinySaana Please open PR with docs update for development what needs to be installed via `apt-get`.

PR #355, opened 2026-03-25
Moonraker filament type parsing fix

gaaat98:
Not sure if this is something that deserves a firmware patch, but it fixes a bug that results in filament type mismatches when there is a space in the filament type provided by the slicer, which in turn blocks filament selection from the Snapmaker U1 print UI.

More details here: https://github.com/Arksine/moonraker/pull/1066

Ideally, Snapmaker could fix it by updating Moonraker once (and if) my fix is merged there, but who knows if that will ever happen.


paxx12:
@gaaat98 Thanks. I will wait for upstream to merge patch, then will pick the patch and maintain it moving forward.

gaaat98:
@paxx12 the developer of moonraker answered my PR and apparently the issue was already addressed in moonraker (https://github.com/Arksine/moonraker/commit/3a54161ce7d274c9dfe70205c2747c97f1eb3294) and it's still present in the U1 because an older version of moonraker is used, it looks like we need to wait for Snapmaker to update it as the fixed version returns JSON instead of plaintext and I cannot find where this output is processed by the firmware :(

paxx12:
I will backport this patch, will ask you to review.

paxx12:
Replaced by https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/355 which does backport from Moonraker.

PR #358, opened 2026-03-29
Backport moonraker metadata: filament_name and filament_type as array

paxx12:
Backports upstream commit 3a54161 to return filament_name and filament_type as JSON arrays for multi-material slices, fixes #355.

Also restructures patch paths under home/lava/moonraker/ to match the target filesystem layout.


paxx12:
@gaaat98 please test this build

gaaat98:
Hi, thanks a lot for the time you are dedicating to this! I am unable to build, but just looking at the changes, I can tell you this will not work because:

1) The `regex_find_strings` helper is not defined in the moonraker version running on the U1, but most importantly
2) I already tried myself porting `regex_find_strings` and the fix you propose here, but in the en,d it does not work because whatever calls metadata.py to parse the metadata cannot process the JSON array produced as output for the filament types, it only seems to be parsing strings separated by ;

With a correct metadata.py that implements `regex_find_strings` you can verify that "filament_type" becomes an array after the patch with this command:
`/usr/bin/python3 /home/lava/moonraker/moonraker/components/file_manager/metadata.py -p /userdata/gcodes -f "<FILENAME.GCODE>"`

We have two options here: the easy one is to take the array and return as ; separated strings or, if you can find where this metadata is processed by the firmware and patch it to parse the JSON array, I tried to locate where this happens but failed miserably.

Let me know what you think, if you can find where to patch I can work on it, or if we decide to go with the easier option, I can update the fix!

PR #389, opened 2026-04-11
Create thermal_optimizations.cfg

SPUNGUH:
Adds TMC2240 driver_slope_control without edits to tmc2240.py. Sets value to 2.

### DIFF
--- a/overlays/firmware-extended/32-feature-klipper-tweaks/root/usr/local/share/firmware-config/tweaks/klipper/thermal_optimizations.cfg
+++ b/overlays/firmware-extended/32-feature-klipper-tweaks/root/usr/local/share/firmware-config/tweaks/klipper/thermal_optimizations.cfg
@@ -1,0 +1,13 @@
+[delayed_gcode set_tmc_slope]
+initial_duration: 2.0
+gcode:
+    # This sends the value 2 directly to the slope_control register 
+    # of the TMC2240 after the machine finishes booting.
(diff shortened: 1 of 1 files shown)


SPUNGUH:
TMC2240 has adjustable driver_slope_control that needs additional enabling in tmc2240.py. This delayed gcode enables the feature and sets it to "2", which is a good middle ground of the 0-3 settings available. This is already apart of mainline klipper and Kalico, and generally you can adjust these values in printer.cfg, when tmc2240.py has the correct edits.

The delayed gcode enables the feature without any edits to tmc2240.py, of which I was unable to have changes stay after editing and reboots.

This change nets a 15-20C drop in X/Y stepper motors via my testing. I have two other printers (QIDI Q2 and QIDI P4) with the driver_slope_control: 2 sets, and no issues, and consistant mid 60C driver temps

I have never done a PR before, but basically for my own implementation I just input the GCODE directly into my printer.cfg, but I assume you can fix this PR after trying it out, and incorperating in the correct place, and making the feature enable/disableble

For reference
https://github.com/Klipper3d/klipper/pull/6771

https://github.com/Klipper3d/klipper/commit/cf3b0475daa0d7154d2f986f94d8c184c7cf39c1#diff-0bd67ba2e312c77323bc71f593ec179b050108ad1387734f1332fe73004d31bd

https://github.com/KalicoCrew/kalico/pull/794/commits/ef566f5ef06c4606bc0eb61a8755db85885a5e30

PR #409, opened 2026-04-20
Add OpenRFID rotating logs and update support docs

paxx12:
Logs are stored and accessible via Fluidd and Mainsail.


benk016:
This version goes back to not allowing the AFC to be tied to a spool in spoolman for auto selection. 

Error evaluating 'gcode_macro SET_SPOOL_ID:gcode': gcode.CommandError: SET_SPOOL_ID is not supported.

PR #417, opened 2026-04-24
Add support for SpoolEase tags

yanshay:
This PR adds support for the SpoolEase encoded tag format.

SpoolEase is a spool inventory management solution (and more) used primarily by Bambu Lab users. Recently, many of these users have also started using Snapmaker U1 printers and expressed interest in having Snapmaker read the SpoolEase tag format directly.

This PR implements the required tag processing for SpoolEase tags. The implementation has been tested successfully by a Snapmaker user using SpoolEase.

It would be great to have this integrated into the extended firmware so other users can benefit from it as well.

### DIFF
--- a/overlays/firmware-extended/13-patch-rfid/root/home/lava/klipper/klippy/extras/filament_protocol_ndef.py
+++ b/overlays/firmware-extended/13-patch-rfid/root/home/lava/klipper/klippy/extras/filament_protocol_ndef.py
@@ -2,12 +2,52 @@ import copy
 import io
 import json
 import logging
+import urllib.parse
 from . import filament_protocol
 
 NDEF_OK = 0
 NDEF_ERR = -1
 NDEF_PARAMETER_ERR = -2
 NDEF_NOT_FOUND_ERR = -3
+
+NDEF_URI_PREFIX_MAP = {
+    0x00: "",
+    0x01: "http://www.",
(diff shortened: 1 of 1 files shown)


virusburger:
I have been working with @yanshay on getting this implemented for the U1. As Yanshay does not have a U1, I have been testing for him on mine. 

So far, I have been using the files in the PR on my U1 without issues for the last week or so. I was able to use a local build of the extended firmware with the required changes. Right now, when a tag is encoded with the SpoolEase tag, the printer will set the filament to the correct vendor and color.

paxx12:
@virusburger @yanshay 

Sorry, but I do not have plans to add additional formats beside the OpenSpool. The current support is good enough for most of use cases. However, you should consider contributing this to `OpenRFID` that this project inherits. This is a proper multi-format solution.

mrgavinconway:
@paxx12 - coming from bambu where spoolease was a real game changer having this available on the snapmaker would ease the transition for some. Might be worth considering again

paxx12:
@mrgavinconway Thanks, but I do not plan to merge additional spool formats parsers to this repository. There's an existing PR open for OpenRFID that appears to be approved. If this project merges this, it will be included in here as well under `OpenRFID`: https://github.com/suchmememanyskill/OpenRFID/pull/13

PR #426, opened 2026-04-29
Add force bed leveling and timelapse print preference tweaks

paxx12:
## Summary

- Adds overlay \`35-feature-print-preferences\` with two new independently toggleable tweaks under **Settings → Tweaks** in the firmware-config UI
- **Force Auto Bed Leveling**: overrides \`SDCARD_PRINT_FILE\` to call \`SET_PRINT_PREFERENCES AUTO_BED_LEVELING=1\` before each SD card print
- **Force Timelapse**: overrides \`SDCARD_PRINT_FILE\` to call \`SET_PRINT_PREFERENCES TIMELAPSE=1\` before each SD card print

## Design

Each cfg is self-contained — enabling a feature symlinks exactly one file. Both cfgs share an identical \`SDCARD_PRINT_FILE\` macro that guards each hook with an existence check before calling, then delegates to the original handler:

\`\`\`
{% if printer["gcode_macro _PRINT_PREF_BED_LEVELING"] %}
    _PRINT_PREF_BED_LEVELING
{% endif %}
{% if printer["gcode_macro _PRINT_PREF_TIMELAPSE"] %}
    _PRINT_PREF_TIMELAPSE
{% endif %}
_SDCARD_PRINT_FILE_BASE {rawparams}
\`\`\`

Each cfg only defines the hook it actually implements. When both are enabled, the last cfg alphabetically (\`timelapse_force.cfg\`) owns \`SDCARD_PRINT_FILE\`; both \`_PRINT_PREF_*\` macros are defined so both preferences fire correctly.

### DIFF
--- a/docs/firmware_config.md
+++ b/docs/firmware_config.md
@@ -49,7 +49,7 @@ Toggle settings directly from the web interface:
 | Klipper Metrics Exporter | Enabled, Disabled | Enable Prometheus metrics |
 | VPN Provider | None, Tailscale | Enable VPN remote access (Experimental) |
 | Cloud | None, OctoEverywhere | Enable Cloud-based remote access (Experimental) |
-| Tweaks | TMC AutoTune, TMC Reduced Current, Object Processing, AFC Stub | Experimental Klipper tweaks ([tweaks](tweaks.md)) |
+| Tweaks | TMC AutoTune, TMC Reduced Current, Object Processing, AFC Stub, Force Auto Bed Leveling, Force Timelapse | Experimental Klipper tweaks ([tweaks](tweaks.md)) |
 | Troubleshooting | Faulty Toolhead Bypass | Temporary toolhead thermistor bypass so the remaining toolheads can still be used ([faulty_toolhead](faulty_toolhead.md)) |
 | RFID Detection System | External, Snapmaker, OpenRFID, OpenRFID (force generic vendor) | Set how filament is detected ([rfid_support](rfid_support.md)) |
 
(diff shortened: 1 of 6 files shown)


MichiMess:
Thanks, I installed the 1.3.0-paxx12-test-pr-426-3209187 firmware, activated it in firmware config, checked the BED_MESH_CALIBRATE ADAPTIVE=1 in the gcode, and checked the Exclude Object in the slicer.

But unfortunately it doesn't work. It just start printing after bed detection.

Am I'm doing something wrong?

Can you tell me how I can test this correctly? I'm using Orca.

Manfredo1952:
> Thanks, I installed the 1.3.0-paxx12-test-pr-426-3209187 firmware, activated it in firmware config, checked the BED_MESH_CALIBRATE ADAPTIVE=1 in the gcode, and checked the Exclude Object in the slicer.
> 
> But unfortunately it doesn't work. It just start printing after bed detection.
> 
> Am I'm doing something wrong?
> 
> Can you tell me how I can test this correctly? I'm using Orca.

Probably not. Same here- Orca  does not listen to the bed mesh command. Klipper reports 
Unknown command:"SET_PRINT_AUTO_BED_LEVELING"

I reverted back to 1.2.0 for now.

MichiMess:
@Manfredo1952 yeah me too

paxx12:
Superseded by https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/429.

PR #429, opened 2026-05-02
Add force bed leveling and timelapse print preference tweaks

paxx12:
Adds `FORCE=1` parameter to `TIMELAPSE_START` and `BED_MESH_CALIBRATE` via klipper patches, allowing macros to bypass the `print_task_config` guards that block these commands when the Snapmaker app has not enabled them.

Each tweak overrides the relevant klipper command to inject `FORCE=1`:

- `timelapse_force.cfg`: overrides `TIMELAPSE_START` with `FORCE=1`
- `bed_leveling_force.cfg`: overrides `BED_MESH_CALIBRATE` with `FORCE=1 ADAPTIVE=1`

Adaptive mesh falls back to full mesh when `exclude_object` is not active.

### DIFF
--- a/overlays/firmware-extended/35-feature-print-preferences/root/usr/local/share/firmware-config/functions/21_settings_tweaks_bed_leveling_force.yaml
+++ b/overlays/firmware-extended/35-feature-print-preferences/root/usr/local/share/firmware-config/functions/21_settings_tweaks_bed_leveling_force.yaml
@@ -1,0 +1,54 @@
+settings:
+  tweaks:
+    items:
+      bed_leveling_force:
+        label: Force Bed Mesh Calibration
+        description: Forces bed mesh calibration before each print.
+        help_url: http://snapmakeru1-extended-firmware.pages.dev/tweaks#force-auto-bed-leveling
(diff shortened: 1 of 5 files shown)


paxx12:
@Manfredo1952 @MichiMess Please test this.

Manfredo1952:
@paxx12   it does work now with OrcaSlicer - I have selected "force adaptive  mesh" in the firmware-config . when printing starts, it still reports unknown commands, but later the adaptive mesh is initiated. 
grafik
So far, so good. Thanks a lot

MichiMess:
> @Manfredo1952 @MichiMess Please test this.

@paxx12, yes  this is working perfectly! Thanks!! Indeed a timelapse error, but timelaps and adaptive mesh runs fine!

image

One last nerdy question, the Cavity Led goes on when the timelaps is starting. With the older firmware I managed to get it always on. I don't know how anymore. Is this still an option with this firmware?

Again, many thanks!

paxx12:
The gcode command is not available, and it will not be. I think Orca has to be updated to actually remove it.

PR #433, opened 2026-05-05
Integrate Spoolman with the NFC filament detection

Rocka84:
# [Feature] Spoolman NFC Integration

## Description
Integrates the U1's filament detection with Moonraker/Spoolman to automatically sync active spools via OpenSpool NFC tags.
Successfully tested on my own machine.

## What's Changed
* Added Web UI toggles under "Spoolman" to independently control NFC sync and built-in macros.
* Extended Klipper's filament protocol and NDEF parser to support `SPOOL_ID`.
* Updated `/printer/filament_detect/set` to handle spool identification from internal and external scanners.
* Added `spoolman_multi_tool.cfg` for seamless multi-tool spool management.

### DIFF
--- a/docs/rfid_support.md
+++ b/docs/rfid_support.md
@@ -101,6 +101,26 @@ Using the non-standard OpenSpool `subtype` field it is possible to specify a mat
 - `additional_color_hexes` - Additional colors for multicolor spools (up to 4)
 - `weight` - Spool weight in grams
 - `diameter` - Filament diameter in mm (e.g., 1.75)
+- `spool_id` - Integer ID for automatic Spoolman synchronization
+
+### Spoolman Integration
+
+The extended firmware optionally integrates with [Spoolman](https://github.com/Donkie/Spoolman) via Moonraker.
+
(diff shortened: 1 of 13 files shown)


paxx12:
@Rocka84 Sorry, but I will not merge support for adding `SPOOL_ID` in this firmware. The problem with `SPOOL_ID` as an identifier encoded in a spool, that it only works with OpenSpool as it requires a specially crafted tag. As such this only works with OpenSpool and cannot properly support any other tags.

Instead I will work on merging that the `CARD_UID` is stored in the Spoolman DB: https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/364. This way the tag does not have to be programmed in order to be attached to Spoolman, as the Spoolman holds all mapping keys. And I believe this is how it should behave. This makes it to support any NFC tag as long as it is detected, it does not even have to be decryptable.

The end user experience is the same: spool is detected and linked to spoolman database, but who holds mapping changes: it is spoolman that holds mapping, not spool.

paxx12:
Here is more explanation for why: https://github.com/paxx12/SnapmakerU1-Extended-Firmware/pull/285#issuecomment-4414700410

PR #439, opened 2026-05-08
Feature: Add G-code MD5 Verification

raistlin7447:
## Summary

Ports the MD5 file-integrity check from [DrA1ex/ff5m](https://github.com/DrA1ex/ff5m) to the U1 extended firmware.

When a g-code file has a `; MD5:<hash>` comment on its first line, Klipper verifies the hash before printing begins. and aborts with a clear error if the file is corrupt or was truncated during upload.  If the file has been corrupted or truncated, Klipper will detect the mismatch, cancel the print, and, by default, delete the file.

## Changes

- **`overlays/firmware-extended/35-feature-gcode-md5/`** — new overlay
  - `gcode_md5.py` — Klipper extras plugin
  - `gcode_md5.cfg` — Klipper config (loaded via extended include via Firmware Config)
  - `35_settings_tweaks_gcode_md5.yaml` — Firmware Config UI toggle
- **`tools/gcode-md5/`** — slicer-side helper scripts
  - `add_md5.sh` — Linux / macOS
  - `add_md5.bat` — Windows
- **`docs/gcode_md5.md`** — user documentation

## Testing

Tested on U1 with extended firmware:

- [X] File without MD5 header prints normally
- [X] File with correct checksum prints normally
- [X] File with corrupted content aborts with "MD5 MISMATCH" error
- [X] Firmware Config toggle enables/disables the check correctly
- [X] `VERIFY_MD5 FILE=...` console command works

## Credit

Original concept and implementation: [DrA1ex/ff5m](https://github.com/DrA1ex/ff5m) (GPLv3).

### DIFF
--- a/docs/firmware_config.md
+++ b/docs/firmware_config.md
@@ -50,6 +50,7 @@ Toggle settings directly from the web interface:
 | VPN Provider | None, Tailscale | Enable VPN remote access (Experimental) |
 | Cloud | None, OctoEverywhere | Enable Cloud-based remote access (Experimental) |
 | Tweaks | TMC AutoTune, TMC Reduced Current, Object Processing, AFC Stub | Experimental Klipper tweaks ([tweaks](tweaks.md)) |
+| G-code MD5 Verification | Enabled, Disabled | Verify MD5 Checksums in G-code files ([gcode_md5](gcode_md5.md)) |
 | Troubleshooting | Faulty Toolhead Bypass | Temporary toolhead thermistor bypass so the remaining toolheads can still be used ([faulty_toolhead](faulty_toolhead.md)) |
 | RFID Detection System | External, Snapmaker, OpenRFID, OpenRFID (force generic vendor) | Set how filament is detected ([rfid_support](rfid_support.md)) |
 
--- a/docs/gcode_md5.md
+++ b/docs/gcode_md5.md
@@ -1,0 +1,116 @@
+---
+title: G-code MD5 Verification
+---
+
+# G-code MD5 Verification
+
+Verifies an MD5 checksum embedded in a g-code file before printing begins.
+If the file has been corrupted or truncated Klipper will detect the mismatch, 
+cancel the print, and report a clear error rather than starting a bad print.
+
+## How it works
+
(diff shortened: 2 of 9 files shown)


paxx12:
It seems that all of this can be implemented by moonraker post processing plugin. I'm unsure why this was made as klipper module.

I also give it very small chance of merging, given how many preconditions this requires. Like configuring slicer. It seems that all of that could be achieved with a post processing script that validates that gcode has `PRINT_END`, so we are sure that file is not truncated as this seems to be a main aim of this change.

raistlin7447:
Thanks for the feedback and really helpful direction.

Truncated uploads seem to be a recurring pain point across many printers and frontends, so I think the underlying feature is worth having even if the implementation needs rethinking.

For the port I followed ff5m's approach since it was a known working implementation, but you're right that Moonraker is a better integration point for this project.

Since all the major frontends, stock screen, Guppy, KlipperScreen, HeliXscreen, route print jobs through Moonraker, a plugin there would catch everything without requiring any slicer configuration.

I'd like to rework this as a Moonraker post-processing plugin built around your `PRINT_END` idea as the core check, with MD5 as an optional layer for users who want stronger integrity verification. That approach would:

- Require zero slicer changes for the base functionality
- Work across all frontends automatically
- Let users opt into MD5 stamping however they prefer: post-processing script or eventually native slicer support (I'll be submitting a PR to OrcaSlicer for native MD5 embedding in the next few days)

Would that direction make this worth reconsidering for a merge?

paxx12:
Yes, since this would be post-processing and optional feature with very minimal surface layer, and also work out of box for most users, I think to have high chance to being merged. Please post proof of concept for me to review ahead of time.

paxx12:
Given the above, I will close this PR for time being. Please open a new one with revisited approach.

paxx12:
@raistlin7447 

This is for example the end of gcode that I see for my silces.

```
; filament end gcode 
PRINT_END
TIMELAPSE_STOP
M73 P100 R0
; EXECUTABLE_BLOCK_END

; CONFIG_BLOCK_START
...
; CONFIG_BLOCK_END
```

PR #451, opened 2026-05-11
Add gcode_integrity_check Moonraker component

raistlin7447:
Follow-up to #439 (closed pending revised approach).

Opening this PR with the new approach for early review per your suggestion. Just the single component file for now. Happy to add config and docs once the approach is approved. Let me know what else you'd like included.

## Approach

`gcode_integrity_check.py` is a Moonraker component that hooks `job_state:started`. As soon as a print begins, it validates the file and issues `CANCEL_PRINT` with the failure reason if a check fails. Configurable defaults that activate out of the box with no user setup.

## PRINT_END check (primary)

Reads the file backwards in 8 KB chunks (constant memory regardless of file size) and scans the last N executable lines for the configured marker. Your example end-of-file works cleanly, same as what I see from Snapmaker Orca and OrcaSlicer::

```gcode
PRINT_END
TIMELAPSE_STOP
M73 P100 R0
; EXECUTABLE_BLOCK_END

; CONFIG_BLOCK_START
...
; CONFIG_BLOCK_END
```

Comment lines match the marker but don't count toward the scan budget, so `PRINT_END` sitting a few executable lines above the tail is found cleanly and the config block trailer doesn't push it out of range.

**Question on default**: `PRINT_END` is in the user-configurable end-gcode section of the slicer, so users who customize it could accidentally break the check. `; EXECUTABLE_BLOCK_END` looks like it's emitted by the slicer itself outside the user's reach. Would that be a better default? Easy to change either way.

## MD5 check (optional, layered)

If the first line is `; MD5:<32-hex-char-hash>`, verifies the hash covers the rest of the file. If the header is absent, the check is skipped.

## On notifications and timing

The check needs to run and fire `CANCEL_PRINT` as fast as possible. Console messages go via Moonraker's websocket rather than Klipper's `RESPOND`, because `RESPOND` enters the gcode queue and would block waiting for homing/probing to complete (~20 s). Pushing via the websocket gets the check running faster and `CANCEL_PRINT` in the queue as soon as practical.

## Known limitation

`CANCEL_PRINT` itself enters the gcode queue and lands after Klipper finishes its current startup sequence (ultimately, the first few lines of the file will execute). So motion doesn't stop instantly. The printer completes its homing pass before the cancel takes effect (~20 s). No damage occurs (no extrusion, no print moves), but the screen returns to idle without a visible reason on it. If this is acceptable, great. If not, the file also contains an optional second handler on `file_manager:filelist_changed` (currently commented out) that validates at upload time and deletes bad files before they can be selected. It's an additional/alternate proof-of-concept.

## Asks

1. Does the approach look right? Especially if `job_state:started` is the correct hook.
2. Acceptable trade-off that the printer's screen doesn't show why a print was cancelled? The alternative is letting bad files print, which seems worse.
3. `PRINT_END` vs `; EXECUTABLE_BLOCK_END` as default marker?
4. Anything specific you'd like in the follow-up config/docs commits?
5. What's the preferred default for the `file_manager:filelist_changed` handler?

   - **(a)** Registered by default alongside `job_state:started` so bad files are caught at upload as well as at print start.
   - **(b)** Gated behind a config option (e.g. `validate_on_upload: False` by default) so users can opt in via `gcode_integrity_check.cfg`..
   - **(c)** Removed entirely.

   Note that the upload-time handler doesn't currently address the no-on-screen-notification problem any better than the print-start path does. The file is deleted and a console alert is emitted, but a user not watching the console will still be left wondering where their file went.

### DIFF
--- a/overlays/firmware-extended/35-feature-gcode-integrity-check/home/lava/moonraker/moonraker/components/gcode_integrity_check.py
+++ b/overlays/firmware-extended/35-feature-gcode-integrity-check/home/lava/moonraker/moonraker/components/gcode_integrity_check.py
@@ -1,0 +1,297 @@
+# gcode_integrity_check.py
+#
+# Moonraker component that verifies gcode file integrity.
+#
+# The primary hook is job_state:started, which validates a file as soon as a print begins. If a check
+# fails, the print is cancelled.
+#
+# An optional second hook on file_manager:filelist_changed can validate files at upload time and delete
+# bad uploads before any print is attempted. It is disabled by default and can be enabled by uncommenting
+# its registration in __init__.
+#
+# Checks (configurable):
+#
+#   End-of-file marker check (default: enabled)
+#     Scans the last N executable lines for a configurable marker. Detects truncated uploads. Default
+#     marker is "PRINT_END"; can be set to any string including a comment such as "; EXECUTABLE_BLOCK_END".
+#
+#   MD5 checksum check (default: enabled)
+#     If the first line is "; MD5:<hash>", verifies the hash covers the rest of the file. Files without
+#     a MD5 header are silently skipped.
+#
+# Both checks fail open: an unreadable file is allowed through. Only a positively detected failure
+# triggers the error action.
+#
+# Configuration (extended/moonraker/gcode_integrity_check.cfg):
+#
+#   [gcode_integrity_check]
+#   require_end_marker: True
+#   end_marker: PRINT_END
+#   executable_scan_lines: 25
+#   enable_md5: True
+#
+# This file may be distributed under the terms of the GNU GPLv3 license
+
+from __future__ import annotations
+
+import hashlib
+import logging
+import os
+from enum import Enum
+from typing import TYPE_CHECKING, List, Optional
+
+if TYPE_CHECKING:
+    from moonraker.confighelper import ConfigHelper
+
+GCODE_EXTENSIONS = (".gcode", ".g", ".gco")
+MD5_HEADER = "; MD5:"
+MD5_HEX_LEN = 32
+
+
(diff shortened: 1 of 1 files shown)


raistlin7447:
Context update that affects the direction here: OrcaSlicer main now natively supports exporting `.gcode.3mf` for third-party (Klipper/Marlin/RRF) printers via a new per-printer `use_3mf` option (https://github.com/OrcaSlicer/OrcaSlicer/pull/14238, merged). That container already embeds a per-plate MD5 of the G-code at `Metadata/plate_N.gcode.md5`, sitting right next to `Metadata/plate_N.gcode`. I verified end-to-end on a current build that the stored digest matches a recomputed MD5 of the embedded gcode, so it's a reliable integrity source straight from the slicer, no post-processing script required.

On the OrcaSlicer side I've closed my feature request (and abandoned my local dev work) for an inline `; MD5:` header (https://github.com/OrcaSlicer/OrcaSlicer/issues/13637) in favor of leaning on this container, since maintaining a parallel in-gcode header isn't worth it when the data already ships in the 3mf.

That raises a design question for this component. Right now it scans the plain gcode tail for a completeness marker. An alternative (or addition) would be to ingest the `.gcode.3mf` directly: unzip, read `Metadata/plate_N.gcode.md5`, and verify it against the extracted gcode before the print starts. The open question is whether the U1's Moonraker upload/print path can accept and unpack a `.gcode.3mf` at all today, or whether that handling would need to be added first. That's the more general path overall: `.gcode.3mf` is becoming the slicer's native delivery format and the container carries more than just the md5 (thumbnails, config, plate metadata), so supporting it sets the firmware up for more than integrity checking alone. Curious how you'd weigh all that before I build anything further.

PR #470, opened 2026-05-22
Fix(AFC): preserve multicolor swatch when assigning Spoolman spool

n19htmares:
When a multicolor filament is configured in U1 (PR #442), assigning a multicolor spool from Spoolman via AFC-Lite was replacing the multicolor swatch on the U1 touchscreen with a single white color (or whatever was primary decided by Spoolman).

Root cause:
_set_spool_from_response() only read color_hex from the Spoolman API response. For multicolor filaments Spoolman returns no color_hex, only multi_color_hexes as a comma-separated string. _set_filament_config() always set COLOR_NUMS = 1 and only populated RGB_1, discarding all additional colors.

Fix:
_set_spool_from_response() now reads multi_color_hexes and passes it to _set_filament_config()
_set_filament_config() accepts a new multi_colors parameter, splits the comma-separated string, sets COLOR_NUMS, MULTI_MODE, and populates RGB_1 through RGB_5 as needed
Falls back correctly to single color behavior when multi_color_hexes is absent

Tested on: Snapmaker U1, Openspool with multi_color_hexes set

Relates to: PR #442 which added multi-RGB infrastructure on the tag write side, this applies the same approach on the Spoolman read side.


n19htmares:
The build failure is pre-existing in the afc-spoolman-auto-register branch and is not related to this PR. 03-add-spool-id-support.patch fails to apply against the v1.4.0 base firmware. This PR only modifies AFC_lane_spoolman.py and the fix is valid — it will apply cleanly once the patch conflict in the base branch is resolved.

PR #474, opened 2026-05-25
Update RFID payload parsing to support FilaMan format

ManuelW77:
Added support for FilaMan tags https://www.filaman.app


paxx12:
@ManuelW77 I'm confused about this PR. It seems that only difference is `"protocol"` but there are no other payload changes. Is that correct? If yes, what is the value of introducing new protocol? In general this repository does not intend to accept additional protocols outside of OpenSpool, I would advise to contribute to https://github.com/suchmememanyskill/OpenRFID.

ManuelW77:
Hmm ok. I wanted to implement the tag payload for FilaMan www.filaman.app into the Snapmaker printer. A lot of people use FilaMan, it's a complete filament manager with own tags and data. The reason for not changing the payload is, that only this data is needed, but the tags have some other data too, so it can't use the OpenSpools Project structure.

paxx12:
@ManuelW77 Thanks. Please open this change to OpenRFID repository. In general, I would advise to provide a distinct implementation, as this allows to treat each format separately without side effects to other one. Whatever lands in OpenRFID will be provided in this project.

ManuelW77:
Ok, i'll do this :)

paxx12:
Per the conversation will close this PR.

PR #480, opened 2026-05-27
feat: Update the OctoEverywhere plugin version

QuinnDamerell:
Hey!

I just recently released `5.0.3`, which has even more perf fixes! It should help the plugin run even more smoothly on lower-power devices like the U1.

Let me know if any changes are needed!

### DIFF
--- a/docs/cloud.md
+++ b/docs/cloud.md
@@ -1,26 +1,25 @@
 ---
-title: Cloud Remote Access (Experimental)
+title: Cloud
 ---
 
-# Cloud Remote Access (Experimental)
+# Cloud
 
-
-Control your printer remotely using cloud-based remote access providers.
+Cloud services provide features like remote access, AI failure detection, notifications, MCP servers, etc.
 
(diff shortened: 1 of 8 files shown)


paxx12:
Thanks @QuinnDamerell. I will merge this in a follow-up. I need to figure out a way to also upgrade existing installs.

QuinnDamerell:
@paxx12 no problem, thanks! I have been working on changes to help with memory pressure, since I know it's an issue for some users. I have a few more changes in the pipeline that will help as well.

Maybe for updating, you can have a simple file somewhere that has a version number in it. Then on boot you can check if the version matches, and if not, do an update. You could then use some enabled feature registry or the existence of file paths to detect what needs to be updated.

For OE, you basically just need to run the install again to pull the new bits and pull any new PY packages. The plugin can be "installed" whenever needed. I use it as a recovery system in my scripts. As long as the data files aren't messed with, the plugin bootstrap still points at them, it will upgrade perfectly

QuinnDamerell:
(Reposted, my phone was logged into the wrong account. 😂)

PR #482, opened 2026-05-27
Feature: Expose Spoolman IDs via HA data bridge macro for Home Assistant tracking

BigMcRat:
Since the switch to AFC-Lite, Spoolman IDs are no longer tracked natively via Spoolman Integration toolhead variables, leaving Home Assistant blind to active spool IDs per Extruder.
This PR introduces a dedicated [gcode_macro HA_SPOOL_ID_PER_EXTRUDER_DATA] block inside afc.cfg to act as a clean data bridge. Moonraker automatically exposes these variables to Home Assistant, allowing users to map active Spoolman IDs per Extruder right alongside their filament sensors from Spoolman. Fully standalone, clean, and tested.


benk016:
I've been trying to accomplish this same thing by manually managing the location tag of the spools with the spoolman integration. This would be awesome to have!

PR #490, opened 2026-05-29
camera: make internal MIPI capture FPS configurable

### LINKED ISSUE #437: paxx12-14 Cam flickering
with Ver. 1.2.0-paxx12-14-71c68d1 the case cam is flickering.

https://github.com/user-attachments/assets/4e81d3d4-9a05-499e-8ad0-dae1e768693a

djgringoboy2003:
The internal MIPI camera capture rate is hardcoded to 30 fps, which beats against 50 Hz mains/LED PWM and produces the visible stream flicker/pulsing reported in #437 (30 fps divides 60 Hz cleanly, but not 50 Hz).

Make the capture rate configurable via a new `camera fps` key, read from extended2.cfg with a default of 30 so existing behaviour is unchanged, and expose a "Capture Frame Rate" selector in firmware-config - Camera offering 30 fps (60 Hz mains) / 25 fps (50 Hz mains). 25 fps divides 50 Hz evenly and removes the flicker for UK/EU/50 Hz users.

- S99v4l2-mpp-mipi: source `camera fps` (default 30) into the capture -fps
- 15_settings_camera.yaml: add camera-fps dropdown; restarts the mipi service
- extended2.cfg seed: document the new key

The USB camera path is unaffected (capture-v4l2-jpeg-mpp sets no -fps).

Closes #437

### DIFF
--- a/overlays/firmware-extended/02-firmware-config/root/usr/local/share/firmware-config/extended/extended2.cfg
+++ b/overlays/firmware-extended/02-firmware-config/root/usr/local/share/firmware-config/extended/extended2.cfg
@@ -15,6 +15,9 @@ internal: paxx12
 usb: none
 # Enable RTSP streaming server: true, false
 rtsp: false
+# Internal (Case) camera capture frame rate: 30 (60 Hz mains), 25 (50 Hz mains)
+# Lower to 25 if the stream flickers/pulses in 50 Hz regions (UK/EU/most of Asia).
+fps: 30
 
 [components]
 rfid: external, snapmaker, openrfid, openrfid-generic
--- a/overlays/firmware-extended/60-app-camera/root/etc/init.d/S99v4l2-mpp-mipi
+++ b/overlays/firmware-extended/60-app-camera/root/etc/init.d/S99v4l2-mpp-mipi
@@ -10,8 +10,11 @@ if [ "$CAMERA_LOGS" = "syslog" ]; then
     CMD_FAKE_SERVICE_ARGS="$CMD_FAKE_SERVICE_ARGS --syslog"
 fi
 RTSP_ENABLED=$(/usr/local/bin/extended-config.py get "$EXTENDED_CFG" camera rtsp false)
+# Capture frame rate. Configurable to avoid flicker from the capture rate
+# beating against mains/LED frequency: 30 for 60 Hz regions, 25 for 50 Hz.
+CAMERA_FPS=$(/usr/local/bin/extended-config.py get "$EXTENDED_CFG" camera fps 30)
 
-CMD_CAPTURE="/usr/local/bin/capture-v4l2-raw-mpp --device /dev/video11 --format nv12 --jpeg-quality 7 --jpeg-sock /tmp/capture-mipi-jpeg.sock --mjpeg-sock /tmp/capture-mipi-mjpeg.sock --h264-sock /tmp/capture-mipi-h264.sock --raw-frame-sock /tmp/capture-mipi-raw.sock --width 1920 --height 1080 --fps 30"
+CMD_CAPTURE="/usr/local/bin/capture-v4l2-raw-mpp --device /dev/video11 --format nv12 --jpeg-quality 7 --jpeg-sock /tmp/capture-mipi-jpeg.sock --mjpeg-sock /tmp/capture-mipi-mjpeg.sock --h264-sock /tmp/capture-mipi-h264.sock --raw-frame-sock /tmp/capture-mipi-raw.sock --width 1920 --height 1080 --fps ${CAMERA_FPS}"
 
 CMD_WEBRTC="/usr/local/bin/stream-webrtc --h264-sock /tmp/capture-mipi-h264.sock --webrtc-sock /tmp/capture-mipi-webrtc.sock"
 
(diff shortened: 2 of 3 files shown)


paxx12:
@djgringoboy2003 I think this is not confirmed solution.

JimDMoran:
This fix also works for my machine, which is on 60hz power in the US.

so, although this change from 30fps to 25fps seems to eliminate the problems of exceedingly bright or flashing video, I believe the root cause is not fully understood.

djgringoboy2003:
I dug into this properly on my own U1 and found the actual mechanism, so I think this can move past "not confirmed" now.

**Root cause: it's a PWM/frame-rate beat, not a mains-frequency thing.**

- The chamber LED (`[led cavity_led]`, `white_pin: PA10`) runs as Klipper software PWM at the *default* `cycle_time: 0.01` → **100 Hz**, confirmed live via Moonraker's `configfile.settings`.
- The internal camera sensor (GC2053) runs at a fixed **33.33 ms frame time (30 fps)**.
- 33.33 ms / 10 ms = **3⅓ PWM cycles per frame**. That fraction is what causes the rolling bands / flicker — each frame samples a different phase of the LED's on/off cycle, so brightness walks frame to frame. At exactly 25 fps, a frame is 40 ms = **4 whole cycles**, so every frame integrates identical light and it's rock solid. This is exactly what @paxx12 originally guessed back in May ("misaligned pwm frequency of led comparatively to capture frequency of camera") — it just needed someone to go measure it.

This is why it doesn't correlate with mains frequency at all — the 100 Hz comes from firmware config, not from AC. That's consistent with JimDMoran seeing it on 60 Hz US power: the PWM/frame beat doesn't care what your wall socket does.

It also explains the "stuck overexposed" variant people are reporting (myself included, and Thierry's latest comment on #437): the camera's AE has a `AecAntiFlicker` mode that normally quantizes exposure to 10 ms steps, which happens to hide the beat most of the time. When AE drifts off that quantization — usually in darker/LED-lit scenes — you get both the flicker *and* AE can mislock on the oscillating brightness, giving the stuck-bright image. That also lines up with why covering the camera or toggling the LED "resets" it temporarily (forces AE to reconverge) and why it's worse at night/on LED-lit shots than in daylight.

**I reproduced this quantitatively:** with AE manually disabled and a fixed sub-cycle exposure, per-frame band brightness variance (std across frames) was ~5.4 at 30 fps / 100 Hz PWM, vs. a ~0.05 noise floor with the LED off entirely. Switching to 25 fps (or to a PWM cycle_time that divides evenly, see below) drops it back to the noise floor.

**Alternative/complementary fix worth considering:** since the actual root cause is the LED PWM frequency, PAXX could instead set `cycle_time: 0.003333` (300 Hz) on `[led cavity_led]` in firmware. 300 Hz divides evenly into both the 30 fps (33.33 ms = 10 cycles) and 25 fps (40 ms = 12 cycles) frame times, so it would fix this for every user without needing a new config option at all. I only tested this optically at a forced short exposure so I can't vouch for MCU PWM load at 300 Hz, but if that's fine on your end it might be the cleaner fix, and it's not mutually exclusive with this PR — happy to have both merged, or just the LED fix if you'd rather not add a camera FPS knob at all.

Either way, I'm confident now this isn't a placebo/workaround — it's a real PWM-vs-frame-rate aliasing bug with a known trigger condition. Let me know if it'd help to have the raw measurement data/script.

djgringoboy2003:
Follow-up with the actual data + a way to reproduce this yourselves, since a couple people (rightly) pointed out a single screenshot can't show a frame-to-frame flicker.

**Reproduction guide, script, and raw data:** https://gist.github.com/djgringoboy2003/775b3e7ef377b46d88e24cfb9904b6de

The gist has:
- `README.md` — full step-by-step: disable AE, force a short exposure, drive the chamber LED via `SET_LED`, capture snapshots, measure, then confirm the fix two ways (25 fps capture, or a 300 Hz `cycle_time` on the LED with no camera change at all). Every command in it uses a placeholder `<PRINTER_IP>` — no hardware/account specifics of mine in there.
- `Measure-CamFlicker.ps1` — the actual measurement script. Pulls a burst of `/webcam/snapshot.jpg`, splits each frame into 27 horizontal bands, and computes each band's brightness variance *across frames* — that's the number that captures flicker, since no single JPEG can show a temporal effect.
- `flicker-results.csv` / `band-profiles.csv` — the raw per-frame numbers behind the stats I posted above.
- `frame-brightness-chart.html` — a small self-contained chart (open it in any browser, no server needed) plotting one image row's brightness across 12 consecutive frames, same conditions except LED PWM frequency:
  - **100 Hz (current default):** sawtooth between ~34 and ~42, alternating almost every frame — this is the flicker.
  - **300 Hz (`cycle_time: 0.003333`):** flat around ~35, same LED duty, same exposure, same 30 fps.

That side-by-side is the clearest way I've found to show this isn't noise or a workaround — it's a deterministic aliasing pattern that appears and disappears exactly when the math predicts it should.

Happy to run any other test someone wants on real hardware — printer's set up for this right now.

paxx12:
@djgringoboy2003 Thank you a lot. I merged this instead https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/574. I hopefully retained your contribution in history of this project.

paxx12:
@djgringoboy2003 I confirmed this with Snapmaker that there's no negative impact on setting 300Hz instead of 100Hz.

PR #445, opened 2026-05-10
Spoolman Idea 2b: Store RFID Tag ID in Spoolman database when selected via AFC-Lite (use Moonraker Agent)

paxx12:
The https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/364 implemented this by adding klipper module. This change instead uses Moonraker Agent to implement this feature avoiding calls from Klipper.

PR #364, opened 2026-03-30
Spoolman Idea 2: Store RFID Tag ID in Spoolman database when selected via AFC-Lite

paxx12:
## Summary

This PR introduces automatic RFID card-to-spool binding for the AFC-Lite system, enabling automatic spool detection and management via Spoolman when an RFID card is present on a lane.

## Key Differences from PR #285

The #285 requires usage of `spool_id` and only works with `OpenSpool` programmed spools. This PR instead stores the RFID to spool mapping in a Spoolman, allowing to support any type of spools as long as they are RFID detected.

## Behavior

This PR does update `lot_nr` of spool in Spoolman depending on the user action:

- It will set `lot_nr` if Spool is assigned to Spoolman via Fluidd (AFC-Lite)
- It will clear `lot_nr` if Spoolman is unassigned
- It will remove `lot_nr` from other spools if spool is assigned to another Spool in Spoolman
- The `lot_nr` will store the `card_uid:XXX,card_uid:YYY` which will be used to map the tag to spool_id each time the spool is inserted into printer
- It only depends on RFID Tag identifier, as such both sides of spool has to be mapped for the first time

PR #163, opened 2026-01-20
Spoolman Helper: a usage tracking module and integration for Spoolman, in U1 flavour

unlucio:
What provides:
- Adds Apple's stuff to .gitignore to prevent Cupertino from soiling the repo
- Adds SPOOL_ID get to filament_protocol_ndef, filament_protocol, and print_task
- Adds a proxy for spoolman's api to moonraker
- spoolman_helper.py: main entry point and orchestrator for the module. It delegates to sub objects for specific behaviours and featurea.
- u1_tools.py: translation interface for teh u1' tool's system
- spoolman.py: kipper client for the spoolman API, it talks to moonraker
- print_lifecycle.py: reacts to a print events
- macros.py_: interface towards the klipper macros world
- logs.py: internal log facility
- commands.py: spoolman helper commands
- all neded cfgs and macros

PR #491, opened 2026-05-29
Add automatic spool detection and Spoolman/SpoolLink integration

paxx12:
## Summary

Automatic spool detection and mapping for the Snapmaker U1 multi-filament
stack, bridging the physical RFID card reader with
[Spoolman](https://github.com/Donkie/Spoolman).

The in-printer web UI has been split out and merged separately as #581
(`Filament Manager`); this PR is the Spoolman/SpoolLink backend it drives.

### Docs

- [Spoolman Integration](https://github.com/paxx12/fork-SnapmakerU1-Extended-Firmware/blob/spoolman-spoollink/docs/spoolman.md) — user guide
- [Design notes](https://github.com/paxx12/fork-SnapmakerU1-Extended-Firmware/blob/spoolman-spoollink/docs/design/spoolman.md) — wire format, custom fields, component flow

### What it does (`38-feature-spoollink`)

SpoolLink is a **Moonraker component** paired with a thin Klipper
`[spoollink]` router. When a channel's RFID card is scanned (or a spool ID
is set from the AFC lane config), it resolves the spool against Spoolman by
ID or card UID, binds the card to that spool, and pushes the filament
metadata back into Klipper. Moonraker's built-in `[spoolman]` active spool
is kept in sync with the toolhead so remaining weight tracks the print. A
local JSON cache keeps things working while Spoolman is unreachable.

The resolution pipeline, custom fields, and wire format are documented in
the design notes above.

### Config-driven enablement

`[spoolman] host` in `extended2.cfg` is the single source of truth. Service
hooks derive the Klipper `[spoollink]` include and the Moonraker
`[spoolman]` / `[spoollink]` blocks from it on start; `firmware-config` only
toggles the `[spoolman]` section.

### AFC wiring

AFC lane status reports a `filament_name` built from vendor, material, and
subtype. Weight is not tracked by AFC-Lite, so `SET_WEIGHT` is a stub.
`06-add-spool-id-support.patch` is rebased for firmware `1.5.1`.

## Why not a `spool_id` on the tag?

An earlier direction was to carry the Spoolman `spool_id` directly on the tag.
That was dropped because it only works with tags written in the OpenSpool
format: the `spool_id` has to be programmed onto the tag, so it excludes every
spool that already ships with an RFID tag that cannot be (re)programmed or uses
a different format. Those spools are still detectable by their hardware UID —
they just can't carry a `spool_id`. It also lets a blank, non-programmed tag be
used as-is: assign it to a spool once and its UID resolves on every later scan,
with nothing written to the tag.

SpoolLink maps the tag UID to a spool inside Spoolman, so it works with any
RFID-detected spool regardless of tag format or writability: the tag stays
read-only, and the card UID ↔ spool association lives in Spoolman's `card_uids`
custom field.

### Prior attempts

- #285 — *Idea 1: add OpenSpool `spool_id` extension* (closed) — required
  specially programmed tags carrying `spool_id`.
- #173 — *NFC tag `spool_id` for Spoolman integration* (closed) — optional
  `spool_id` field parsed from OpenSpool JSON.
- #163 — *Spoolman Helper* (closed) — parsed `SPOOL_ID` off the tag as well,
  bundled with a full Spoolman usage-tracking module (Moonraker proxy,
  print-lifecycle tracking, macros).
- #364 — *Idea 2: store the RFID tag ID in Spoolman* (closed) — the pivot to
  UID-based mapping (via `lot_nr`), superseding #285.
- #445 — *Idea 2b: same, via a Moonraker agent* (closed).

This PR continues the #364/#445 direction, moving the mapping to the dedicated
`card_uids` custom field and the SpoolLink Moonraker component.

## Warning

The format where the spool mapping is stored changed in this PR https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/364! It is no longer stored in the `lot_nr` field, instead a new special `extra.card_uid` field was introduced. This means that all current mappings are incorrect.

You can convert those manually via Spoolman. Enable this PR, and then manually move the data from `Lot Nr` to `Extra Fields > Card UID`, removing the `card_uid:` prefix:

image

### DIFF
--- a/docs/afc-lite.md
+++ b/docs/afc-lite.md
@@ -76,3 +76,7 @@ rm /oem/printer_data/config/extended/klipper/afc.cfg
 
 ![AFC Tools Remapping](images/afc_tools.gif)
 
+## See Also
+
+- [Spoolman Integration](spoolman.md) — automatic spool lookup, card UID binding, and active spool tracking via the SpoolLink component
+
--- a/docs/design/spoolman.md
+++ b/docs/design/spoolman.md
@@ -1,0 +1,169 @@
+---
+title: Spoolman Integration Design
+---
+
+# Spoolman Integration Design
+
+# Spoolman Custom Fields
+
+| Entity | Key | Purpose |
+|--------|-----|---------|
+| `spool` | `card_uids` | Comma-separated uppercase hex NFC UIDs associated with the spool |
+| `filament` | `variant` | Filament subtype / variant (e.g. "Silk", "Matte") |
+
+Both fields use Spoolman's double-serialised string convention: the inner value is
+JSON-encoded before storage, so a single UID `"AABBCCDD"` is stored as `"\"AABBCCDD\""`.
+`spoollink` decodes these on read and re-encodes on write.
+
+## `card_uids` Custom Field
+
+NFC tag UIDs are stored in a Spoolman custom field named `card_uids` on the spool entity.
+
+### Field definition (created on first connection test)
+
+```json
+{
+  "key": "card_uids",
+  "name": "Card UIDs",
+  "entity_type": "spool",
+  "field_type": "text",
+  "order": 1,
+  "default_value": "\"\""
+}
+```
+
+### Wire format
+
+The `extra.card_uids` value is a **JSON-encoded string** — the string value itself is
+wrapped in JSON quotes by the client before sending, so it arrives as a JSON string
+containing another JSON string:
+
+```
+extra.card_uids = "\"AABBCCDD,11223344\""
+```
+
+When decoded, the inner string is a comma-separated list of uppercase hex UIDs:
+
+```
+AABBCCDD,11223344
+```
+
+### Encoding rule
+
+Before writing, the comma-separated UID string must be JSON-encoded:
+
+```
+jsonEncode("AABBCCDD,11223344") → "\"AABBCCDD,11223344\""
+```
+
+### Decoding rule
+
+On read, strip the outer JSON quotes if present; handle both encoded and raw forms:
+
+```
+raw = "\"AABBCCDD,11223344\""
+decoded = AABBCCDD,11223344
+uids = ["AABBCCDD", "11223344"]
+```
+
+### PATCH body for `card_uids`
+
+```json
+{
+  "extra": {
+    "card_uids": "\"AABBCCDD,11223344\""
+  }
+}
+```
+
+---
+
+## UID Format
+
+NFC tag UIDs are formatted as **uppercase hex with no separators**:
+
+```
+AABBCCDD        (4-byte UID)
+04A1B2C3D4E5F6  (7-byte UID)
+```
+
+---
+
+## Sync Logic ("Add to current, remove from others")
+
+When an NFC tag is scanned or assigned to a spool:
+
+1. **Fetch** the target spool via `GET api/v1/spool/{id}`.
+2. **Append** the tag UID to `extra.card_uids` if not already present.
+3. **Write** the updated UID list via `PATCH api/v1/spool/{id}`.
+4. **Search** for other spools that contain the same UID by fetching all spools
+   (`limit=1000&allow_archived=true`) and filtering client-side on `card_uids`.
+5. **Remove** the UID from each other spool's `card_uids` and write back via `PATCH`.
+
+---
+
+## `variant` Custom Field (Filament)
+
+The filament variant (e.g. "Silk", "Matte") is stored in a Spoolman custom field named
+`variant` on the filament entity.
+
+### Field definition (created on first connection test)
+
+```json
+{
+  "key": "variant",
+  "name": "Variant",
+  "entity_type": "filament",
+  "field_type": "text",
+  "order": 1,
+  "default_value": "\"\""
+}
+```
+
+The value follows the same JSON-encoded string convention as `card_uids`:
+
+```
+extra.variant = "\"Silk\""
+```
+
+### Create Filament body with variant
+
+```json
+{
+  "name": "Galaxy Black Silk",
+  "material": "PLA",
+  "extra": {
+    "variant": "\"Silk\""
+  }
+}
+```
+
+# `spoollink` Component Flow
+
+`spoollink` is a Moonraker component (`moonraker/components/spoollink.py`), loaded when a
+`[spoollink]` section is present in the Moonraker config. It runs in Moonraker's event loop
+and uses the built-in `http_client`, `klippy_apis`, and `spoolman` components directly, so
+it does not maintain its own WebSocket connection.
+
+1. On construction, `spoollink` calls `server.register_remote_method("spoollink_resolve_spool", ...)`;
+   Moonraker re-registers it with Klipper on every Klippy `ready`, so the binding survives
+   Klipper restarts.
+2. On startup (`component_init`), `spoollink` calls `GET api/v1/field/spool` and
+   `GET api/v1/field/filament` to verify the `card_uids` and `variant` custom fields exist,
+   creating them via `POST api/v1/field/{entity}/{key}` if missing.
+3. Klipper (or AFC) calls `spoollink_resolve_spool` with `channel`, `spool_id`, and/or `card_uid`.
+4. `spoollink` resolves the spool from Spoolman:
+   - By ID: `GET api/v1/spool/{id}`
+   - By card UID: `GET api/v1/spool?limit=1000`, then filter client-side
+     on `extra.card_uids` (comma-separated, JSON-encoded uppercase hex UIDs).
+5. If both `spool_id` and `card_uid` are given and the card is not yet in the spool's
+   `card_uids`, `spoollink` appends it via `PATCH api/v1/spool/{id}` with
+   `{"extra": {"card_uids": "\"UID1,UID2\""}}` (JSON-encoded string, as required by
+   Spoolman's custom field API). Any other spool that already carries the same UID has it
+   removed via a subsequent `PATCH` — a UID can only belong to one spool at a time.
+6. On success, `spoollink` pushes the resolved filament info into Klipper by calling the
+   `spoollink/set` endpoint (registered by the Klipper `[spoollink]` router), which merges
+   it onto `filament_protocol.FILAMENT_INFO_STRUCT` and applies it to `print_task_config`.
+7. Klipper stores the metadata in `print_task_config` and notifies subscribers.
+   `AFC_lane.get_status()` surfaces `spool_id` to Fluidd/Mainsail.
+
--- a/docs/index.md
+++ b/docs/index.md
@@ -63,6 +63,7 @@ Heavily expanded firmware with extensive features and customization:
 - [Faulty Toolhead Bypass](faulty_toolhead.md) - Temporary bypass for one failed toolhead thermistor so the other toolheads can still be used
 - [RFID Filament Tag Support](rfid_support.md) - NTAG213/215/216 support for OpenSpool format
 - [Alternative Filament Detection](rfid_support.md#alternative-detection-systems) - Alternative detection implementations with extended spool/tag support from Bambu, Creality, Anycubic, and others
+- [Spoolman Integration](spoolman.md) - Automatic filament metadata sync and spool tracking via Spoolman, resolving spools by ID or RFID card UID
 
 **Monitoring & Notifications:**
 
--- a/docs/spoolman.md
+++ b/docs/spoolman.md
@@ -1,0 +1,50 @@
+---
+title: Spoolman Integration
+---
+
+# Spoolman Integration
+
+Automatic filament metadata sync and spool tracking via
+[Spoolman](https://github.com/Donkie/Spoolman).
+
+## What It Provides
+
+- Resolves a Spoolman spool by ID or RFID card UID and applies its
+  metadata (vendor, material, variant, colour) to the extruder channel.
+- Associates RFID card UIDs with spools automatically, so scanning a
+  known card loads its filament without any manual step.
+- Tracks the active spool in Moonraker so Spoolman can update remaining
+  filament weight as you print.
+
+## Enabling
+
+Enable via Fluidd/Mainsail settings under
+**Snapmaker Components > Spoolman Integration**, set the Spoolman host,
+and reboot. Set the same toggle to **Disabled** to turn it off.
+
+## GCode Commands
+
+### `SET_SPOOL_ID`
+
+Assign a Spoolman spool to an AFC lane. Reads the lane's RFID card UID,
+binds it to the spool (so a later scan resolves automatically), and
+applies the filament metadata to the channel:
+
+```
+SET_SPOOL_ID LANE=E0 SPOOL_ID=5
+```
+
+| Parameter | Default | Description |
+|-----------|---------|-------------|
+| `LANE` | — | AFC lane name (e.g. `E0`) |
+| `SPOOL_ID` | `0` | Spoolman spool ID; `0` clears the assignment |
+
+## Limitations
+
+- Spoolman must be reachable from the printer over HTTP.
+- Variant defaults to `Basic` for Snapmaker-branded filaments when not
+  set in Spoolman; empty for all other vendors.
+
+For the wire format, custom fields, and component flow see the
+[design notes](design/spoolman.md). For AFC lane status that surfaces
+`spool_id` see [AFC-Lite](afc-lite.md).
--- a/overlays/firmware-extended/02-firmware-config/root/usr/local/share/firmware-config/extended/extended2.cfg
+++ b/overlays/firmware-extended/02-firmware-config/root/usr/local/share/firmware-config/extended/extended2.cfg
@@ -37,3 +37,7 @@ cloud: none
 [monitoring]
 # Enable Klipper Prometheus exporter on specified address
 # klipper_exporter: :9101
+
+# Spoolman server URL for filament tracking; managed via firmware-config.
+# Example: host: http://192.168.1.100:7912
+[spoolman]
--- a/overlays/firmware-extended/13-patch-rfid/patches/06-add-spool-id-support.patch
+++ b/overlays/firmware-extended/13-patch-rfid/patches/06-add-spool-id-support.patch
@@ -1,0 +1,59 @@
+--- a/home/lava/klipper/klippy/extras/print_task_config.py
++++ b/home/lava/klipper/klippy/extras/print_task_config.py
+@@ -24,6 +24,7 @@
+     'filament_edit': [True] * PHYSICAL_EXTRUDER_NUM,
+     'filament_exist': [False] * PHYSICAL_EXTRUDER_NUM,
+     'filament_soft': [False] * PHYSICAL_EXTRUDER_NUM,
++    'filament_spool_id': [0] * PHYSICAL_EXTRUDER_NUM,
+     'extruder_map_table': [i for i in range(PHYSICAL_EXTRUDER_NUM)] + [0] * (LOGICAL_EXTRUDER_NUM - PHYSICAL_EXTRUDER_NUM),
+     'extruders_used' : [False] * PHYSICAL_EXTRUDER_NUM,
+     'extruders_replenished': [i for i in range(PHYSICAL_EXTRUDER_NUM)],
+@@ -181,6 +182,7 @@
+
+             tmp_print_task_config['filament_official'][channel] = info['OFFICIAL']
+             tmp_print_task_config['filament_sku'][channel] = info['SKU']
++            tmp_print_task_config['filament_spool_id'][channel] = info.get('SPOOL_ID', 0)
+             if self.filament_param_obj is not None:
+                 tmp_print_task_config['filament_soft'][channel] = \
+                     self.filament_param_obj.get_is_soft(info['VENDOR'], info['MAIN_TYPE'], info['SUB_TYPE'])
+@@ -282,7 +284,10 @@
+             allowed_edit = False
+             if self.print_task_config['filament_exist'][ch]:
+                 if self.print_task_config['filament_official'][ch] == False:
+-                    allowed_edit = True
++                    if (self.print_task_config['filament_spool_id'][ch] or 0) > 0 and self.printer.lookup_object('webhooks').has_remote_method('spoolman_set_active_spool'):
++                        allowed_edit = False
++                    else:
++                        allowed_edit = True
+
+             tmp_filament_edit[ch] = allowed_edit
+         if tmp_filament_edit != self.print_task_config['filament_edit']:
+@@ -343,6 +347,7 @@
+         filament_soft = gcmd.get_int('SOFT', None)
+         filament_color = gcmd.get_int('FILAMENT_COLOR', None)
+         filament_color_rgba = gcmd.get('FILAMENT_COLOR_RGBA', None)
++        filament_spool_id = gcmd.get_int('FILAMENT_SPOOL_ID', None)
+         filament_alpha = gcmd.get_int('ALPHA', None, minval=0, maxval=255)
+         filament_color_nums = gcmd.get_int('COLOR_NUMS', None, minval=1, maxval=FILAMENT_COLOR_NUMS_MAX)
+         filament_colors_str = gcmd.get('COLORS', None)
+@@ -359,6 +365,8 @@
+
+             if tmp_print_task_config['filament_official'][config_extruder] and bool(force) == False:
+                 raise gcmd.error("[print_task_config] filament_config, official filament, not configurable!")
++            if (tmp_print_task_config['filament_spool_id'][config_extruder] or 0) > 0 and not force and filament_spool_id == None and self.printer.lookup_object('webhooks').has_remote_method('spoolman_set_active_spool'):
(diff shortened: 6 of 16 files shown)


MKloberg:
Some good news: I have 3 U1's running on this build now, in production with zero issues.
I migrated all my spoolman spool UID's over to the new extra field in the new format before I started deployment of the new image. I did find some issues with my own tags, but that had nothing to do with this PR.

Converting the 3rd printer just now, it totally worked.
What I saw across all of them was that it simply "forgot" the spoolman URL that I'm 100% sure it knew before the upgrade.
Yet, a vanity issue.

After loading random filament into unit 3, it didn't know what it was at first but after pressing the "Read Spool Tags" button, all four slots lit up with the correct correlations to Spoolman. 
This route is going very well, at least for me.

image

MKloberg:
This is also stress testing my farm and 3DMRP, matching this approach for filament identification and tracking as we print.
I don't know what else to say, it's working now after sitting on the sidelines for a few weeks while we had this debate on how to identify spools.

image

MKloberg:
6420f23 still good, running in production with zero issues. Everything is working as expected with normal printing on 2 U1's.
Running a massive order, tomorrow I plan to onboard a 3rd U1 and do some Hueforge paintings which will bring the focus back to tagging and running previously unknown filaments.

benk016:
Noted on Discord, but updating here as well. This PR solves the issue of filament loading clearing spoolman data, as long as the spools are tagged with the new method.

Daroonn:
I have trouble reading cheap tags (NTAG215). They lack a digital signature. Has anyone noticed a similar problem in themselves?

benk016:
I am just using the cheapest NTAG215 tags I could find on Amazon without issue. 

I have had maybe 3 that refused to write so far. I have about 25 left of the initial 100 I bought.

BigMcRat:
6420f23 It has also been running smoothly for me for 3 days.

Daroonn:
I solved the tag problem. It turned out that the filament type cannot contain any special characters. I had a "PLA+" and that was a problem. Now everything works without any problem.

MKloberg:
e49a032 works great. I'll upgrade another one next and will report back.
This is what it looks like on my dashboard mid print (which means the AFC lanes are tracking and responding nicely, even over API calls and the webhooks). Spools are NFC tagged on both sides and I didn't have to correct anything manually during filament loading, which is a first. Just grab a spool and stick it in there, yay!!!

image

MKloberg:
I upgraded another one and that worked just as well. Note here that in lane 0, is a gradient filament - I was surprised to see that show up correctly on the touch screen, very nice. Awesome work!!!

image

MKloberg:
Running in production with e49a032 now on 3 U1's with no issues.
Where this is still a little wobbly is that there seems to be no way to tell if the filament is just pre-loaded (the feeder fed it up to the toolhead), but not actually in the toolhead yet.
By "telling" I mean, you can't tell visibly in Fluidd or from the api exposed states of the afc lanes or anywhere else if a toolhead is pre-loaded or fully loaded.
You can see it in the operator panel screen, but that's the only place (I assume it knows, by the state of the filament sensor inside of the extruder).
Perhaps I missed it, but couldn't we use the filament sensor state on each toolhead and then build the logic on top of that?

Case in point, couple of times now, I accidentally just pre-loaded filament, launched the job and then it just stopped with a filament anomaly, causing me to cancel the job/back to the filament touch screen, load that lane and then back to the job to start it. Not a big deal, just wondering if this could be made better even further.

benk016:
That is interesting behavior. When I've pre-loaded filament but didn't "load" it and then start a print, it just loads the filament automatically and then starts printing. I've done it several times.

MKloberg:
> That is interesting behavior. When I've pre-loaded filament but didn't "load" it and then start a print, it just loads the filament automatically and then starts printing. I've done it several times.

In which way? Was it a re-print from the op-panel, an upload from the slicer or via 3DMRP/moonraker api?

benk016:
From the slicer for sure, I think it also worked from the panel on the printer but I can't remember for sure. 

I don't think any of my jobs from 3DMRP have been in that scenario. 

I'll try all three ways to verify. I don't think I've hit this since being on the latest release yet

MKloberg:
Well, my main point was actually that we have currently no external visibility on that filament presence sensor inside the toolhead as far as I can tell (I may be wrong).
Also, there is some confusion about what "loaded" actually means, as there are inherently two states in the workflow of loading that the Afc lanes do not properly reflect.

I'm starting to suspect that there is an auto load setting somewhere (I'm beginning to recall seeing that), but can't find it atm. This may just be on for you and still off for me.

benk016:
When sending a print from the slicer or starting from the printer screen UI, it does a pre-loading sequence in the console as "Set action code PRINT_AUTO_FEEDING"

```
22:33:31 
// Success: Changed main state to PRINTING with action IDLE
22:33:31 
// Success: Set action code PRINT_SWITCH_CHECKING
22:33:31 
// extruder3 -> extruder3
22:33:31 
// Extruder extruder3 already active
22:33:31 
// Success: Set action code IDLE
22:33:32 
// Success: Set action code PRINT_AUTO_FEEDING
22:33:32 
// extruder3 -> extruder3
22:33:32 
// Extruder extruder3 already active
22:34:50 
// Success: Set action code IDLE
22:34:50 
$ ; SpoolLink: E4 no spool found for card 0452287DCD2A81
22:34:50 
// extruder3 -> extruder3
22:34:50 
// Extruder extruder3 already active
```

And the UI shows this 
image

When printing from 3DMRP, it does not do that part and just skips into probing. 

```
10:49:52 PM 
// Success: Changed main state to PRINTING with action IDLE
10:49:52 PM 
// Success: Set action code PRINT_SWITCH_CHECKING
10:49:52 PM 
// Success: Set action code IDLE
10:49:52 PM 
// extruder -> extruder
10:49:52 PM 
// Extruder extruder already active
10:49:58 PM 
// z offset: -0.05
10:49:58 PM 
// sample 2, discard 0
10:50:01 PM 
// probe at x: 140.000, y: 140.000 is z=303.664167
10:50:04 PM 
// probe at x: 140.000, y: 140.000 is z=303.662500
10:50:04 PM 
// Result is z=303.663333
10:50:06 PM 
// probe at x: 140.000, y: 140.000 is z=303.664167
10:50:06 PM 
// sample 2, discard 0
10:50:09 PM 
// probe at x: 140.000, y: 140.000 is z=303.663333
10:50:11 PM 
// probe at x: 140.000, y: 140.000 is z=303.654167
10:50:11 PM 
// Result is z=303.658750
10:50:11 PM 
// z_mesh_complete: -0.009158530743671702
10:50:12 PM 
// Success: Set action code DETECT_PLATE
```

benk016:
Also looks like the run-out sensors are triggered when it's not fully loaded as well. 
Screenshot_20260609_231119_Chrome

ChrisBGL:
I flashed the current PR build and tested it with the internal U1 RFID readers. The basic workflow works well.

My use case is an external RFID reader setup, because my spools are stored in filament dryers and cannot reliably be read by the internal U1 readers.

Would it be possible to expose a supported way to pass an externally read UID to SpoolLink, e.g. something like:

```gcode
RESOLVE_SPOOL_UID LANE=E0 CARD_UID=56844DCE
```

That would allow external readers such as snapmaker-u1-remote-rfid-reader / CheapSpoolDisplay to only send `lane + UID` to the U1, while SpoolLink keeps the Spoolman lookup and filament config logic centralized.

paxx12:
The external reader should use existing api. It will auto resolve spoolman
correctly then. It needs to send card uid.

ChrisBGL:
Thanks for the quick response.

I tested it:

With the filament on the internal RFID reader, it works perfectly and the spool is correctly resolved from Spoolman:
image


However, when I use this API call:

```bash
curl -X POST "http://192.168.178.94/printer/filament_detect/set" -H "Content-Type: application/json" -d "{\"channel\":0,\"info\":{\"CARD_UID\":[66,213,72,237]}}"
```

the card UID is shown correctly in the UI, but the spool is not resolved from Spoolman:
image


One thing I noticed:
With the internal RFID reader, the UID is displayed without `:` separators, while with the API call it is displayed as `42:D5:48:ED`.

Maybe the normalization/matching path is different between internal RFID reads and external API calls?

paxx12:
Someone building the external RFID already tested and confirmed it working. Check `spoollink.log`. However, in this case those are two different UIDs.

ChrisBGL:
Sorry, wrong Screenshot.

Internal: 
image

External:
image

But after reboot it works 👍Thank you!

paxx12:
The screenshot shows blank data. So, it does work or not? :)

ChrisBGL:
It works! 👍

sven-reichelt:
What type of filament detection should be used (if using the internal sensors)?

image

shivalwolf:
I have been running this build for the last few days and have not noticed any issues with my own tags or the original snapmaker one

sven-reichelt:
> I have been running this build for the last few days and have not noticed any issues with my own tags or the original snapmaker one

I've been running build version 387023b since June 15 2026, and have printed for over 150 hours with various filaments since then—including Snapmaker filaments and my own custom tags (for Sunlu and Tinmorry). It's working perfectly for me! I'm curious to see when this feature will be officially available :)

ni4223:
Saw some issue recently where it was not mapping spoolman spool properly to the extruder, even though i could see all the spoolman spool in AFC and rfid manager, used some of AI magic to debug, here is the summary of my session, see if it helps

## Repeating `printer.spoollink.set` → `-32601 Method not found`

  **Repro:**
  1. Running this PR with Spoolman enabled — working fine.
  2. Flashed a `1.5.x` test build (no spoollink feature).
  3. Switched back to this PR.

  After step 3, every card scan logged:

  JSON-RPC Request Error - Requested Method: printer.spoollink.set, Code: -32601, Message: Method not found

  **Cause:** the two halves came back out of sync after the firmware swap.

  - The Spoolman moonraker cfg persists across versions, so the daemon starts and calls `printer.spoollink.set`.
  - But the Klipper `[spoollink]` object wasn't loaded (missing `extended/klipper/spoollink.cfg` symlink), so the endpoint was never registered.

  Confirmed via `/printer/objects/list` — no `spoollink` object, even though the daemon looked healthy (`spoollink_resolve_spool` was registered, but that's the daemon's *inbound* callback, unrelated to the endpoint).

  **Fix:** disabling → re-enabling Spoolman in the firmware-config UI recreated the symlink, restarted everything in order, and the errors stopped.

  **Suggestion:** have the daemon check the endpoint exists before calling it, and log one clear message like `Klipper [spoollink] not loaded — is Spoolman enabled on this firmware?` instead of spamming `-32601` on every scan.

paxx12:
@ni4223 This is valid. Once you flash another firmware, the symlink will be removed. Unsure yet, how to solve that. But will keep it in mind.

I'm likely slowly going into having a hidden auto-generated config on klipper start that is always included, and is templated.

Tom950:
Tried this latest build (5b0c50c) several times but klipper fails to start on reboot. 

Screenshot_20260713_190459_Chrome

convicte:
This is perhaps a very low importance question, so please excuse me if it is.

Trying to understand the intention of the current interface, so that I act accordingly for my `spoolman` instance.

The existing `Filament Manager` is offering the filament `subtype` record. 
When it specifically comes to `PLA - Matte` it's a material but it's also a `Subtype` for all other materials, so you could effectively end up with a material `PLA - Matte` that is also a `subtype: Matte` or it's a material `PETG` with  a `subtype: Matte`

image



Second idiosyncracy are custom `subtypes` - good example being Sunlu filament where they go completely silly with `2.0` but also `2.0 Plus` and also `Plus`. Is it intended to add `Custom` to `Subtype` just like for the `Brand`?

Finally, how should we prep the Spoolman library of filament to play nice with this and prevent mostly all the tags from being in a Mismatch state.  

image


At this time the materials such as `Matte, Plus, 2.0 Plus` etc. are dash separated from the generic name, so perhaps I need to add a `subtype` customer field for them to be recognized correctly?

image

paxx12:
> This is perhaps a very low importance question, so please excuse me if it is.

Sure. Good question. The Snapmaker considers two parameters: Material and Subtype/Variant. Subtype/Variant serves like a modifier (Matte, Basic, SuperSpeed, etc.). What you see in your case like Material `PLA - Matte` is what was imported from your Spoolman. It is expected that Material is primary type, like `PLA`, without modifier as shown in Tag Mismatch.

In general the `PLA - Matte` as material is not understood by Snapmaker firmware, and by Snapmaker Orca following this PR and Spoolman/Spoollink integration. The Variant has to be stored as dedicated field. In general, the best that Spools are uploaded using dedicated iOS/Android apps: https://develop.snapmakeru1-extended-firmware.pages.dev/spoolman#apps. I would propose to use SpoolKid or SpoolPainter.

convicte:
> > This is perhaps a very low importance question, so please excuse me if it is.
> 
> Sure. Good question. 

Thank you very much for the clarification!

Since my Spoolman predates my U1 by several years I still have roughly 60 filaments being in various states of misalignment with best practices.

Now if I understand correctly the best for direct recognition is a custom `variant` field in Spoolman and if I understand correctly I should add any `variant` name I find on the spool - such as all the funny HF+, 2.0 Plus, Extra, Super, etc. into Spoolman to be recognized correctly both up and downstream - from Spoolman to Orca to U1 firmware?

Finally, I have found `Spool Studio` (https://github.com/GeorgHo/SpoolStudio)to be much more feature rich (capable of updating information in both directions - up to `Spoolman` and down to `U1` as well as duplicating spools, creating new once and updating their tags), so was hoping to use it instead of Spool Painter, but perhaps there are some conflicts there with Spool UID and Variants utilization

Would you advise against it?

Thank you once again for all your effort and fantastic contributions to the U1 ecosystem!

paxx12:
@convicte Just run the this branch, this should create `variant` field on `Filaments`. Iterate through `Filaments`, update `Material` to be `PLA/PETG/etc.` and add `Variant` for cases where you need `Basic/Matte` or something different. The next time you load spool all should be good for you. You can ignore the mismatch, but I assume most of them will be fixed.

paxx12:
@convicte Try this on one spool first. You can force reload of Spoolman information with `Refresh` button to see exactly what was loaded.

ni4223:
This SpoolStudio is new, Is it designed to work with this UID mapping flow?

Also surprised to see dev credited my app.  lol
> Original project: [SpoolPainter by ni4223](https://github.com/ni4223/SpoolPainter)

Edit: Jus remembered someone did reach out to me on email and i agreed, but didnt know what became of it, but that was v1 version of my app.

convicte:
Understood!
Already running `1.5.2-paxx12-test-pr-567` and testing all possible scenarios.
Will test the `variants` asap!

Can you please comment in any way on the `Spool Studio` utilisation in your firmware-RFID ecosystem?

Thanks

paxx12:
> Can you please comment in any way on the Spool Studio utilisation in your firmware-RFID ecosystem?

I have not tested if it supports the https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/blob/develop/docs/design/spoolman.md.

convicte:
> This SpoolStudio is new, Is it designed to work with this UID mapping flow?
> 
> Also surprised to see dev credited my app.  lol
> > Original project: [SpoolPainter by ni4223](https://github.com/ni4223/SpoolPainter)
> 
> 

The previous `v1.0 branch` looked exactly like yours until the latest 2.0 version which deviated from it a lot, hence the attribution.

It specifically calls out 

`Printer mapping is designed for:

Snapmaker U1
paxx12 Extended Firmware
Moonraker
Spoolman integration
`

With the latest version focusing on material information:
https://github.com/GeorgHo/SpoolStudio/releases/tag/v2.0.4

Perhaps @GeorgHo could speak to this and update his code to the latest UID flow.

ni4223:
> The previous `v1.0 branch` looked exactly like yours until the latest 2.0 version which deviated from it a lot, hence the attribution.

Looks like dev hasnt kept up with the latest UID flow, so it wont integrate with the this upcoming release, spoolstudio is still using `lot_nr`

This is based on quick code lookup i did
```
 suspend fun findSpoolByLotNr(lotNr: String, forceRefresh: Boolean = false): FilamentSpool? 
```

For now your best bet is one of the app in the doc Paxx linked above

convicte:
> For now your best bet is one of the app in the doc Paxx linked above

Left the author an issue to consider aligning with the new UID flow - perhaps he will find it acceptable.

GeorgHo:
Hi, thanks for documenting the new SpoolLink / UID flow here.

I maintain Spool Studio and have reviewed this PR. The app currently still uses the legacy mapping flow, so I will update it to support the new Paxx12 SpoolLink integration while keeping a legacy fallback for older firmware versions.

The planned changes are tracked here:
https://github.com/GeorgHo/SpoolStudio/issues/1

Planned direction:
- detect the new SpoolLink mode automatically,
- use `SET_SPOOL_ID LANE=E0..E3 SPOOL_ID=<id>` for mapping,
- read `print_task_config.filament_spool_id`,
- support Spoolman's `extra.card_uids` field for RFID/NFC UID mapping,
- add a guided migration path for older UID mappings if needed.

I will test this against a suitable v1.5.2 / SpoolLink build before publishing the compatible Spool Studio release.

paxx12:
There was no official release with lot nr. I would not advise to support
legacy path.

convicte:
> There was no official release with lot nr. I would not advise to support
> legacy path.

If I may add, adding a reliable and robust 'branch' detection mechanism is likely to absorb quite a lot of resources in both creation and testing so if that is not entirely necessary per Paxx12, you could consider a clean cut and the new flow, as have happened in this project.

Food for though!

ni4223:
@convicte  since you plan to test the whole flow, could you test that using SpoolPainter v2.3?
I released it in open testing recently and want to merge it to prod, but because i dont have access to my U1 and my tags i havent been able to test all flows.

Github release: https://github.com/ni4223/SpoolPainter/releases/tag/v2.3.0
Play store open testing: https://play.google.com/apps/testing/com.spoolpainter.app

LMK what you think and if its stable enough for the release

convicte:
Sure- give me a day or two and I'll let you know.

convicte:
@ni4223 come to think of it, I am also in the closed testing group from the start so inadvertently have been running 2.3 for some time now (since it was published for testers)

I must admit I didn't realise and I am not sure what specifically you'd like deep tested for this release?

LMK if there are any particular tricky areas and I can focus on these.

Cheers

ni4223:
> @ni4223 come to think of it, I am also in the closed testing group from the start so inadvertently have been running 2.3 for some time now (since it was published for testers)
> 
> I must admit I didn't realise and I am not sure what specifically you'd like deep tested for this release?
> 
> LMK if there are any particular tricky areas and I can focus on these.
> 
> Cheers

I'm mainly looking for an overall sanity check.

The last release only adds two small features:

> * Type to search the pickers. The Spool and Filament pickers now have a search box at the top. Start typing to filter by material, brand, colour, name, or Spoolman ID instead of scrolling a long inventory. Material / Brand / Colour pickers stay scroll-only.
> * Closest-match suggestions on tag read. When you read a tag that is not yet paired (a vendor tag, or an OpenSpool tag with no matching UID), the closest-matching filaments float to the top of the pickers, best match first, so you can confirm a link in one tap. Matching considers material, brand, colour, and variant. Selecting a filament also floats its own spools. Nothing is auto-selected; the reorder happens only when you open a picker.

I was able to test both features individually, but I didn't get a chance to do a proper end-to-end test.

So if you could give it a shot and let me know how it feels overall, that'd be great. I don't know if you've used v2 of the app before, so a fresh set of eyes would actually be really helpful.

Thanks for your help

convicte:
> So if you could give it a shot and let me know how it feels overall, that'd be great. I don't know if you've used v2 of the app before, so a fresh set of eyes would actually be really helpful.


OK, so I've done the following - up and downstream flow test for the tags (n=10) and Spoolman for the moment - without involving U1 in it at all.

1) Read existing tags previously created with Spool Painter and Spool Studio
2) See if they are recognized to within specific spool in Spoolman, with correct filament/type/color
3) Adjust filament in Spool Painter and push up via Update in Spoolman for color, variant, material
4) Select spool from Spoolman menu in the app and write to tag
5) Read the re-writen tag to confirm it's recognized inside Spool Painter to match Spoolman filament spool number.

Results:
1) Both were recognized, getting matched correctly to spool# while 2 only populating fields, but failed to match a correct spool, but more on this later.
2) Consistently after reading and/or writing a tag 1-2 times, the app would stop recognizing the presence of the TAG near the reader coil and fail to read. I know this is not the issue with the tag since the phone would know the tag is there (vibrate gently when it got near the reader coil) but the app would fail to recognize it. As soon as I closed and reopened the app it would read just fine - this happened 5-6 times so it should be easily reproducible.
3) Reading a tag which matches to a specific spool in Spoolman followed by the Spool being updated in Spoolman doesn't get recognized in the app (even when I pool down to refresh). I needed to deselect the spool in the app and select it again from the search box or close the app and reread the tag. Without it it would say, for example PLA even though Spoolman had it edited to PETG and no matter how many times I pulled down to refresh, it would stay PLA. 
4) This is more of a 'user story' rather than a bug, but it creates a bit of a buggy experience - the app doesn't have a way to handle colorless or multi-color spools. With a clear/transparent/dual/tricolor etc. the whole flow completely breaks down, since you need to select a color from the short list or from a color picker to save a spool - but obviously it will mess up your Spoolman record. 
5) The allocation of `variable: variant` for material subtype works just fine and so does the UID. Changing the subtype updates Spoolman `Variant` fine. What feels like an intentional omission is the ability to edit the material form the APP. The material is greyed out and can't be changed to push this up to Spoolman, which I am deducing is some limitation you encountered?

That's that for the moment. 
Will play around with the flow down to U1 and Orca later today or tomorrow.

Hope this helps!

ni4223:
@convicte Thanks so much for the detailed testing, this is exactly the kind of end-to-end pass I was hoping for, and every one of these is useful. This is firt time these are reported, so this is very helpful.

  Quick notes on each:

Recognized/matched correctly — great to hear. For the 2 that populated fields but didn't match a spool, if you can tell me anything about those tags vendor tag? written by Spool Studio? anything different about them), that'll help me chase it.

 Stops reading after 1-2 taps until you reopen the app — this is the big one, and reproducible is gold. The phone buzzing but the app not seeing the tag tells me the tag's fine, the app just isn't armed to receive it, so it's an NFC dispatch issue on my end. I've got a good idea where it is and I'll dig in. If you happen to notice whether it's tied to something specific (right after a write, or after opening the color/camera picker), that'd help narrow it.

Spoolman edit doesn't show after refresh — dug into this one. When you select a spool, the app takes a one-time snapshot of its data into the form. Pull-to-refresh re-fetches your inventory in the background, but it doesn't re-apply that fresh data to the spool you already have selected, so the form keeps showing the old snapshot (your PLA vs PETG). On top of that, re-picking the same spool is currently a no-op, which is why only deselecting + reselecting or closing the app works. Clear bug, I'll fix the refresh path.

Colorless / multi-color spools — you're right, this is a known gap. The app models one color per spool today, so clear/transparent and dual/tri-color don't fit, you're forced to pick a single hex that then doesn't represent the real spool. Multi-color hex is already on my list. 

Material greyed out — that one's intentional, not a limitation I hit. On an existing spool, changing material or brand almost always means "wrong filament picked" rather than "this filament is now a different material," so I lock them to avoid corrupting the record; the intended path is to pick the correct filament instead. A proper "re-pair to a different filament" flow is on the roadmap. Curious though, were you editing material to fix a mistake, or genuinely repurposing a spool? That'd help me design it.

  Looking forward to the U1 + Orca results whenever you get to them. Thanks again, seriously helpful.
  
  Also we can take this to my repo so we do not keep messing this thread which is completely unrelated.
  https://github.com/ni4223/SpoolPainter/issues/6
  
  Please use this issue if you have more info. Thanks again

GeorgHo:
Short update from the Spool Studio side:

Spool Studio v3.0.0 beta is now available with support for the new Paxx12 SpoolLink / UID based workflow:

https://github.com/GeorgHo/SpoolStudio/releases/tag/v3.0.0-beta

The beta no longer writes the old spool-id based tag format. It stores the physical NFC card UID in Spoolman, supports conversion of older Spool Studio tags after confirmation, and includes checks for the required Spoolman / printer integration fields.

Tested with Paxx12 Extended Firmware v1.5.2-paxx12-21 beta / PR #567. Feedback from real-world testing is very welcome.

paxx12:
@GeorgHo This change is released. If you would mind opening PR and adding your Ko-Fi or BMC link to docs/: https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/releases/tag/v1.5.2-paxx12-21

GeorgHo:
Thanks, v3.0.2 is now released and tested with v1.5.2-paxx12-21.

Release:
https://github.com/GeorgHo/SpoolStudio/releases/tag/v3.0.2

I opened a small docs PR to link the current Spool Studio v3 release from the Spoolman integration app table:
https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/625

The support link column is intentionally left unchanged for now. I’ll add a Ko-fi / Buy Me a Coffee link in a separate follow-up once it is set up.

PR #496, opened 2026-05-31
(patch) Add AHT2x/AHT3x sensor support and temperature_combined fix

jimmyjon711:
Main Changes:
- Bringing in mainline klipper AHT10 changes to work with AHT20 sensors, this also allows ViViD internal temperature sensors to work.
- Adding `set_read_tolerance`  method so that temperature_combined can be used which is currently needed for ViViD units.

Test:
- Manually compiled image and tested on my U1 that has a ViViD using the AHT30 sensor and EMU using the AHT20 sensor, both work correctly on my machine. 

image

@paxx12 If you would like some sort of documentation added for this, please let me know where you would like it.

### DIFF
--- a/overlays/firmware-extended/11-patch-klipper/patches/home/lava/klipper/08_temperature_combined.patch
+++ b/overlays/firmware-extended/11-patch-klipper/patches/home/lava/klipper/08_temperature_combined.patch
@@ -1,0 +1,27 @@
+From 62351025f7a8bde0de99bc5b1ea3964d6d4f816b Mon Sep 17 00:00:00 2001
+From: jimmyjon711 <jcmadill1@gmail.com>
+Date: Sat, 30 May 2026 13:45:04 -0700
+Subject: [PATCH 1/1] Adding set_read_tolerance method to temperature_combined
+ file so that users can combine sensors to become one
+
+---
+ klippy/extras/temperature_combined.py | 3 +++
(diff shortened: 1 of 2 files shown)


paxx12:
@jimmyjon711 Can you reference me the upstream changes?

overlays/firmware-extended/11-patch-klipper/patches/home/lava/klipper/08_temperature_combined.patch
```diff
@@ -0,0 +1,27 @@
+From 62351025f7a8bde0de99bc5b1ea3964d6d4f816b Mon Sep 17 00:00:00 2001
```
paxx12, line 1: @jimmyjon711 Would you mind pushing this to upstream klipper?

paxx12 (approved):

overlays/firmware-extended/11-patch-klipper/patches/home/lava/klipper/08_temperature_combined.patch
jimmyjon711: This is a fix only for U1 Klipper since the `set_read_tolerance` method calls is only in U1 Klipper and not in mainline/upstream Klipper.

jimmyjon711:
I retested this branch with my changes locally, everything still works and looks good.

PR #568, opened 2026-07-09
Add St0rmingBr4in mod: firmware-config add-on installers

St0rmingBr4in:
## Summary

Personal mod (per `docs/mods.md`) adding an **Add-ons** settings group to the firmware-config web UI, with Installed / Not Installed toggles for optional software that is intentionally not baked into the image. Each toggle runs an idempotent on-device install recipe (all support `--uninstall`), with output streamed to the browser:

- **HelixScreen** — replaces the stock touchscreen UI via the official installer (native Klipper touchscreen UI with a dedicated Snapmaker U1 backend). Ships `S99fb-http-restore` so HelixScreen's boot hook does not silently break the extended firmware's Remote Screen feature (`/screen/`) — its autostart replaces `S99fb-http` with a delegate that never starts the fb-http daemon.
- **Git** — Debian aarch64 binary + git-core helpers + the `libcurl-gnutls.so.4` ABI shim needed by `git-remote-https`. Useful for AFC's `update-afc.sh` and GitOps workflows.
- **Spoolman** — SQLite backend, init script, and Moonraker `[spoolman]` wiring. The spool database lives in `printer_data` and survives uninstall/reinstall.

Also adds an **Overlay Persistence** setting (Settings > System) managing the `/oem/.debug` flag: the recipes hard-require it and fail fast with a pointer to the toggle instead of silently producing installs that vanish on reboot.

The settings follow the existing firmware-config conventions: `get_cmd` derives the current value from on-disk state, per-option `confirm` dialogs, no state stored anywhere else.

## Testing

- `sh -n` on all recipes; YAML validated against `firmware-config.py`'s `deep_merge`/`handle_get_settings` logic (merged cleanly with all existing function files, no group collisions).
- Full firmware build with the mod applied: `./dev.sh make build PROFILE=extended-St0rmingBr4in` completes and the resulting image contains all mod files with correct modes.
- Not yet flashed to a printer; marked draft until validated on-device.


St0rmingBr4in:
On-device validation done (ssh deploy onto a running printer, no reflash). Live testing caught three real issues, fixed in 817c6cf:

- BusyBox `wget` has no TLS — recipes now download via the firmware's bundled `/usr/local/bin/curl` and export it onto the non-interactive `PATH` (the firmware-config daemon's env doesn't include `/usr/local/bin`).
- HelixScreen's U1 platform hooks default to the **DRM backend**, which bypasses `/dev/fb0` — the Remote Screen mirror stays frozen on the splash forever. The recipe now pins `HELIX_DISPLAY_BACKEND=fbdev` in `helixscreen.env`.
- Its evdev touch fallback probes `/dev/input/event1`; the U1's tlsc6x touch is `event0`. Pinned via `HELIX_TOUCH_DEVICE`.

Verified working end to end: install from the settings toggle path, HelixScreen UI rendering through `/screen/` with touch injection, `get_cmd` state detection, and the fb-http restart fixup. One caveat: `S99fb-http-restore` only takes effect at boot when shipped in the squashfs (busybox freezes the boot glob before the overlay pivot), i.e. after flashing a build with this mod — same constraint HelixScreen's own installer documents for its hooks.

overlays/mods/St0rmingBr4in/afc-u1-bay-prep.patch
jimmyjon711, line 1: I don't think it's a good idea to push patches for AFC this way into the extended firmware.

PR #571, opened 2026-07-12
Add Firmware Config web shortcut and build concurrency

### LINKED ISSUE #570: Add Firmware Config shortcut button to web UI
## Summary
Add a small, non-invasive shortcut button in Fluidd/Mainsail that links users directly to the Extended Firmware Config page at /firmware-config/.

## Motivation
The Firmware Config UI is useful, but users currently need to know or remember the /firmware-config/ URL. A visible shortcut in the web UI would make the feature easier to discover without changing the firmware-config service itself.

## Proposed approach
- Serve a small static JavaScript file, for example /firmware-config-shortcut.js.
- Inject a deferred script tag into Fluidd/Mainsail index.html during firmware build.
- The script adds a floating Firmware Config link/button pointing to /firmware-config/.
- Keep it as a standalone overlay so it is easy to review and can fail harmlessly if the frontend layout changes.

## Scope
This issue is only for the Firmware Config shortcut button.

It is not related to Anycubic ACE support, RFID changes, or any other experimental integration work.

## Acceptance criteria
- Button appears in the web UI after building/flashing extended firmware.
- Button opens /firmware-config/.
- Fluidd/Mainsail still load normally if the shortcut script fails to load.
- Implementation is isolated from unrelated feature work.

Tareku99:
## Summary
- Add a standalone overlay that serves `/firmware-config-shortcut.js`.
- Inject a small floating `Firmware Config` shortcut into Fluidd and Mainsail during firmware build.
- Link the shortcut to `/firmware-config/`.
- Collapse the shortcut to a compact tab after a short delay, then expand on pointer hover or keyboard focus.
- Add workflow concurrency so newer pushes cancel older in-progress builds for the same workflow/ref.

## Scope
This PR is focused on issue #570 and the build workflow concurrency helper.

image
image

Visual example of what this looks like

## Checks
- `git diff --check`
- `bash -n overlays/firmware-extended/69-app-firmware-config-shortcut/scripts/01-inject-shortcut.sh`

Closes #570


Tareku99:
Been using this build and everything is working fine, not sure what all the requirements are to make the cut in the next release. This is not a new feature, just a UX improvement imo.
I tried to make it as non-intrusive and no maintenance as possible.

Edit: please let me know about the workflow concurrency change, from personal experience, it is a nice edit to have when many people are pushing fast changes to a branch. Saves on time and money :)

Tareku99:
Pulled in develop after the recent commits, tested again and all is working well. 
@paxx12 Ready for review

paxx12:
@Tareku99 This is very nice change. Thank you for updating it.

Do you have suggestion how we would support other pages, like Filament? I'm kind of thinking that the best is if we could contribute both to fluidd and mainsail ability to add custom iframes in form of pages, and in form of small cards to avoid abusing Cameras :)

Tareku99:
@paxx12 Thanks for taking a look!

I dug a bit more into what you were referring to. 
I think the clean long-term solution is an upstream feature in both Fluidd and Mainsail:

- Native custom application pages in the sidebar/navigation.
- Optional small dashboard cards backed by an iframe.
- A common configuration format, possibly building on the existing .theme/navi.json approach.

I found these related Fluidd efforts, although they may be stale:

- https://github.com/fluidd-core/fluidd/pull/1802 (supersede by PR 1786)
- https://github.com/fluidd-core/fluidd/pull/1786

I think if PR 1786 gets implemented, we should have a solid ground to add onto for what we need on fluidd.

As far as Mainsail goes, it already seems to have a useful starting point through its .theme/navi.json navigation entries, which could expose Filament Manager as a normal sidebar page. However, its current iframe support is still part of the webcam service:

- https://github.com/orgs/mainsail-crew/discussions/2387

That does not solve the dashboard-card use case, though. Ideally, the navigation/plugin model could be extended with a first-class custom-page and optional iframe-card type, shared consistently between Mainsail and Fluidd.

Do we want to add something on our side to make it easier or should we open an issue on Mainsail/Fluidd and get that going?

PR #583, opened 2026-07-15
Add HelixScreen install action to Firmware Config menu

### LINKED ISSUE #582: Add HelixScreen install option to Firmware Config menu
﻿## Summary

Add a convenience action in the Firmware Config web UI to install HelixScreen (an alternative touchscreen UI for the Snapmaker U1) via the existing one-liner from https://helixscreen.org/installation/#snapmaker-u1.

This follows the same pattern as the existing Tailscale and OctoEverywhere install actions. The YAML function simply runs:

    curl -sSL https://releases.helixscreen.org/install.sh | sh

No HelixScreen code is bundled in this repo. The action just triggers their remote installer script, same as we do for other third-party tools.

HelixScreen's developer docs describe overlay-based integration as the cleanest deployment strategy (Level 3: Extended Firmware Overlay), so they're aligned with this approach.

Scope:
- One new YAML function file under overlays/firmware-extended/ (pattern: NN_actions_helixscreen.yaml)
- Optional status item to show if HelixScreen is installed
- Optional Restore Stock UI action for rollback

For discussion:
- Settings or Actions?
- Any concerns about resource usage or conflicts?

Tareku99:
Adds a convenience action in the Firmware Config web UI to install [HelixScreen](https://helixscreen.org/), an alternative touchscreen UI for the Snapmaker U1.

**What this adds** (single overlay: \70-app-helixscreen/\):

| Feature | Description |
|---------|-------------|
| **Status** | Settings > Snapmaker Components shows whether HelixScreen is installed at \/userdata/helixscreen\ |
| **Install** | One-click install button that runs \curl -sSL https://releases.helixscreen.org/install.sh \| sh\ |
| **Restore Stock UI** | Actions > System button to remove HelixScreen and re-enable the stock touchscreen |

No HelixScreen source code is bundled. This follows the same pattern as Tailscale and OctoEverywhere — a YAML function file that triggers a remote installer.

Closes #582

### DIFF
--- a/overlays/firmware-extended/70-app-helixscreen/root/usr/local/share/firmware-config/functions/70_settings_helixscreen.yaml
+++ b/overlays/firmware-extended/70-app-helixscreen/root/usr/local/share/firmware-config/functions/70_settings_helixscreen.yaml
@@ -1,0 +1,49 @@
+settings:
+  snapmaker_components:
+    items:
+      helixscreen:
+        label: HelixScreen
+        description: Alternative touchscreen UI for the Snapmaker U1 by Preston Brown.
+        help_url: https://helixscreen.org/
+        get_cmd:
+          - bash
+          - -c
+          - test ! -x /usr/bin/gui && echo "installed" || echo "not_installed"
+        options:
+          installed:
+            label: Installed
+            confirm: "Remove HelixScreen and restore the stock touchscreen interface?"
+            cmd:
+              - bash
+              - -ec
+              - |
+                chmod +x /usr/bin/gui 2>/dev/null || true
+                rm -rf /userdata/helixscreen
+                /sbin/reboot
+          not_installed:
+            label: Not Installed
+            cmd:
+              - bash
+              - -ec
+              - |
+                echo ">> Installing HelixScreen..."
+                /usr/local/bin/curl -sSL https://releases.helixscreen.org/install.sh | sh
+                echo
+                echo ">> HelixScreen installed. The screen will switch to HelixScreen."
+        default: not_installed
+
+actions:
+  system:
+    items:
+      helixscreen-update:
+        label: Update HelixScreen
+        description: Check for and apply the latest version of HelixScreen.
+        if_cmd: test ! -x /usr/bin/gui
+        cmd:
+          - bash
+          - -ec
+          - |
+            echo ">> Updating HelixScreen..."
+            /usr/local/bin/curl -sSL https://releases.helixscreen.org/install.sh | sh
+            echo
+            echo ">> HelixScreen updated."
--- a/overlays/firmware-extended/70-app-helixscreen/root/usr/local/share/firmware-config/functions/70_settings_helixscreen.yaml
+++ b/overlays/firmware-extended/70-app-helixscreen/root/usr/local/share/firmware-config/functions/70_settings_helixscreen.yaml
@@ -8,7 +8,7 @@ settings:
         get_cmd:
           - bash
           - -c
-          - test ! -x /usr/bin/gui && echo "installed" || echo "not_installed"
+          - test -d /userdata/helixscreen && test ! -x /usr/bin/gui && echo "installed" || echo "not_installed"
         options:
           installed:
             label: Installed
@@ -38,7 +38,7 @@ actions:
       helixscreen-update:
         label: Update HelixScreen
         description: Check for and apply the latest version of HelixScreen.
-        if_cmd: test ! -x /usr/bin/gui
+        if_cmd: test -d /userdata/helixscreen && test ! -x /usr/bin/gui
         cmd:
           - bash
           - -ec
--- a/overlays/firmware-extended/70-app-helixscreen/root/usr/local/share/firmware-config/functions/70_settings_helixscreen.yaml
+++ b/overlays/firmware-extended/70-app-helixscreen/root/usr/local/share/firmware-config/functions/70_settings_helixscreen.yaml
@@ -12,16 +12,6 @@ settings:
         options:
           enabled:
             label: Enabled
-            confirm: "Remove HelixScreen and restore the stock touchscreen interface?"
-            cmd:
-              - bash
-              - -ec
-              - |
-                chmod +x /usr/bin/gui 2>/dev/null || true
-                rm -rf /userdata/helixscreen
-                /sbin/reboot
-          disabled:
-            label: Disabled
             cmd:
               - bash
               - -ec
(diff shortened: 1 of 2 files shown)


Tareku99:
Tested and everything seems to be working well.
@paxx12 Ready for review

overlays/firmware-extended/70-app-helixscreen/root/usr/local/share/firmware-config/functions/70_settings_helixscreen.yaml
```diff
@@ -0,0 +1,50 @@
...
+          - test -d /userdata/helixscreen && test ! -x /usr/bin/gui && echo "enabled" || echo "disabled"
+        options:
+          enabled:
+            label: Enabled
+            cmd:
+              - bash
+              - -ec
+              - |
+                echo ">> Installing HelixScreen..."
+                /usr/local/bin/curl -sSL https://releases.helixscreen.org/install.sh | sh
```
paxx12, line 20: Sorry, but I will not accept unsolicated `install.sh` from helixscreen. This repository follows SBOM, as such all dependencies needs to be versioned and natively integrated. There were cases of helixscreen breaking user printers, as it was not adapted to work with specific versions of the firmware.

paxx12 (changes requested):

overlays/firmware-extended/70-app-helixscreen/root/usr/local/share/firmware-config/functions/70_settings_helixscreen.yaml
justinh-rahb: @paxx12 I can relay feedback to the helixscreen project if there is a preferred upstream method of installation that should be adopted for the U1.

paxx12: @justinh-rahb I don't think this is needed. They publish assets already that can be pulled and mixed in. I will definitely not use `install.sh` as it is simply too messy. There's no need though for this.

Tareku99: I believe that currently the helix is tall scripts checks fkr compatibility before installing. 
From my understating, it also only disables the built in GUI and installs his own Ina different folder which gets nuked with each firmware flashes. 

Thought it was rather safe, altought we cannot check for a nasty push that would be done to the install script (but at the same time, most users don't check the install script anyway so it would not effect the current user behavior imo)

paxx12: Please, no. This is impossible to validate. The `install.sh` is 330kB... script. There were already enough of reports people breaking their U1 with `install.sh`, due to incompatibility with new firmware.

Tareku99: > @justinh-rahb I don't think this is needed. They publish assets already that can be pulled and mixed in. I will definitely not use `install.sh` as it is simply too messy. There's no need though for this.

If that is the direction, it can be closed. 

Altought, don't we have a similar behavior with octoprint and other components?

paxx12: No. Strict versioning for Tailscale and OctoEverywhere.

Tareku99: > Please, no. This is impossible to validate. The `install.sh` is 330kB... script. There were already enough of reports people breaking their U1 with `install.sh`, due to incompatibility with new firmware.

For sure, I do not wish to go against the vision so whatever is the preferred way. I am perfectly okay with it. Not attached to the work :)

paxx12: @Tareku99 

Ask claude to review a bit:

```
- ensure_lmd_running restarts lmd whenever it finds it stopped — exactly true while lmd's own init script is mid-start. I proved it: with the upstream function, a second lmd spawns and would collide with S90lmd's own start-stop-daemon; without it, none.
- start_remote_screen exists only because their installer hijacks S99fb-http as a boot hook. We don't touch it, so 61-app-remote-screen's S99fb-http still runs on its own.
- ensure_wifi_associated is the genuine gap. I grepped the firmware: nothing else brings up wlan0 from the saved config — the stock gui really does own association, reading /oem/printer_data/gui/wpa_supplicant.conf and pushing it into wpa_supplicant. So a network set up on the stock screen won't auto-rejoin; users re-join from HelixScreen's own WiFi panel. I documented that as a known limitation — say the word if you'd rather I port those ~15 lines inline.
```

This is for example what the platform hooks of helixscreen do. This is actually pretty bad how many hooks/and hacks are done to make it work.

paxx12:
Replaced by: https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/590/changes.

overlays/firmware-extended/70-app-helixscreen/root/usr/local/share/firmware-config/functions/70_settings_helixscreen.yaml
Tareku99: Yea doesn't look great. I will review it better for my own sake when I get time later today. 

I would assume that regardless of what changes could be made in helixscreen's code, we are still not willing to include this. 

Edit: I see it being closed, sorry for the waste of time, hard to grasp what is in scope and what is not sometimes!

paxx12: Take a look at https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/590/changes. It is nowhere finished, but is rather slim change, pulling minimal required.

Tareku99: This looks like a much safer implementation of what I was trying to do. Tried looking at it on my phone but I understand the approach.

I assume as more releases of extended firmware, helix screen will be re-avaluated and bumped if needed? Will move convo to the new pr. Leaving this stale.

PR #585, opened 2026-07-15
Allow to upgrade firmware using `.zip`

### LINKED ISSUE #584: Reject non-.bin firmware uploads early with clear error
## Problem

Uploading a non-firmware file (e.g. a `.zip`) through the Firmware Config web UI's upgrade upload handler completes the full upload before failing with a cryptic error:

```
Error ret = -14, please enter 'upfileUnpack -h' to view help information.
[UPG] upgrade_all: Unpack upgrade file failed. Upgrade abort.
```

This is confusing — the user has to wait for the entire file to transfer before discovering it was the wrong format.

## Solution

Validate the uploaded file's extension immediately after the multipart upload completes, before invoking `systemUpgrade.sh`. If the original filename doesn't end in `.bin`, reject with a clear message and clean up the temp file.

### Changes

`stream_multipart_to_file()` now extracts and returns the original filename from the `Content-Disposition` header alongside the file path and size. `handle_upgrade_upload()` checks the extension and returns an early 400 error if it's not `.bin`.

`overlays/firmware-extended/02-firmware-config/root/usr/local/bin/firmware-config.py`

Tareku99:
Fixes #584

## Problem

Uploading a non-firmware file (e.g. a .zip) through the Firmware Config upgrade UI would complete the full upload, then fail with a cryptic `upfileUnpack` error -14. Same for URL downloads pointing at zipped firmware artifacts.

## Solution

Add ZIP unpack support to `30_upgrade.yaml` for both upgrade flows, so .zip firmware files work the same as .bin files.

### Changes

`overlays/firmware-extended/02-firmware-config/root/usr/local/share/firmware-config/functions/30_upgrade.yaml`

Both `upgrade_url` and `upgrade_upload` actions now:

1. **Detect ZIP files** — checks the first 2 bytes for `PK` magic to identify zip archives (works regardless of file extension)
2. **Extract** — unzips to `/userdata/.tmp_upgrade/zip_unpack` (avoids RAM-backed `/tmp` limits), finds the first .bin inside (case-insensitive)
3. **Size sanity check** — rejects files under 50 MB before calling `systemUpgrade.sh`
4. **Clear errors** — fails early with a readable message if unzip fails or no .bin is found

### Behavior

| Scenario | Before | After |
|---|---|---|
| Upload .bin | Works | Works |
| Upload .zip with .bin inside | Fails with error -14 | Extracts and upgrades |
| URL to .bin | Works | Works |
| URL to .zip artifact | Fails with error -14 | Downloads, extracts, upgrades |
| ZIP missing .bin | Fails with error -14 | Early error: "No .bin file found in the zip archive" |
| Tiny download (error page) | Fails with error -14 | Early error: file too small |

### DIFF
--- a/overlays/firmware-extended/02-firmware-config/root/usr/local/share/firmware-config/functions/30_upgrade.yaml
+++ b/overlays/firmware-extended/02-firmware-config/root/usr/local/share/firmware-config/functions/30_upgrade.yaml
@@ -8,12 +8,6 @@ upgrade_url:
         echo "Please download the artifact manually."
         exit 1
         ;;
-      *.bin) ;;
-      *)
-        echo "WARNING: URL does not end with .bin - firmware files should be .bin files."
-        echo "Continuing in 5 seconds..."
-        sleep 5
-        ;;
     esac
 
     echo "Downloading upgrade..."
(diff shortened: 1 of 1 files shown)


overlays/firmware-extended/02-firmware-config/root/usr/local/bin/firmware-config.py
```diff
@@ -640,11 +647,20 @@ def handle_upgrade_upload(self):
                 self.send_error(400, "Expected multipart/form-data")
                 return
 
-            file_path, file_size = self.stream_multipart_to_file("file", upload_path)
+            file_path, file_size, orig_name = self.stream_multipart_to_file("file", upload_path)
             if not file_path:
                 self.send_error(400, "No file in request")
                 return
 
+            if orig_name and not orig_name.endswith(".bin"):
```
paxx12: What if the upgrade script would support .zip? I'm more inclined to simply reject this via `.yaml` by checking the file extension.

I'm more inclined to extend the upgrade script to support `.zip` and automatically unpack the file.

paxx12 (changes requested):

overlays/firmware-extended/02-firmware-config/root/usr/local/bin/firmware-config.py
Tareku99: I thought about it but chose against it. I think it is okay to front load the unpacking to the user, avoids zip related malicious attacks that extended firmware shouldn't need to worry about. 
These changes were based on the fail early philosophy, I wasn't necessary adding features but if you think it is valuable to include, I can work on it.

justinh-rahb: > What if the upgrade script would support .zip? I'm more inclined to simply reject this via `.yaml` by checking the file extension.
> 
> I'm more inclined to extend the upgrade script to support `.zip` and automatically unpack the file.

I was going to chime in to say it would be better to support .zip insead of rejecting it (early or late).

Tareku99: If that is what is wanted I can push a change and add that! 
Shall I do that?

paxx12: Yup. Only `30_upgrade.yaml` change should be required.

Tareku99: Sounds good, will do that when done with my work day

Tareku99:
`=== Upgrade Started ===
/userdata/web_upgrade.bin
Size: 225783792 bytes
Time: 2026-07-16 01:11:51
========================================

Detected .zip archive, extracting...
Extracted firmware from archive.
Starting upgrade (238 MB)...
Name                State    Build time      Version
Main processor      N/A      20260709180031  1.5.2.10
Main MCU            online   20260709172742  5bf934edda
Print head0 MCU     online   20260709172742  5bf934edda
Print head1 MCU     online   20260709172742  5bf934edda
Print head2 MCU     online   20260709172742  5bf934edda
Print head3 MCU     online   20260709172742  5bf934edda
[UPG] additional_operations_before_active: Additional operations before activating the soc firmware.`

seems like it is working @paxx12

paxx12 (approved):

PR #587, opened 2026-07-15
Update firmware to `1.5.2`

paxx12:
Upstream changelog:

- Fixed an issue introduced in V1.5.0 where, in some cases, the first
  layer was overly squished when Heated Bed Leveling was not enabled
  before printing.
- Fixed an issue where the nozzle diameter could be incorrect in some
  cases.

Note: V1.5.2 is a Beta maintenance release for V1.5.0 and V1.5.1, and
provides full support for the official Top Cover. Top Cover features
also require Snapmaker Orca V2.3.5 or Snapmaker App V2.3.5.

Note: upstream's `_upgrade_firmware_desc.json` for this release still
carries the V1.5.0 `name`, `version`, `size`, `sha256` and release
notes, so it cannot be used to verify the binary. `FIRMWARE_SHA256`
was computed from the published `_upgrade.bin` (221912136 bytes)
instead, and is not corroborated by an upstream checksum.

Note: after installing, re-run Multi-Toolhead Offset Calibration with
no filament loaded, and enable Heated Bed Leveling before printing.


erazorlll:
Seems Snapmaker integrated an additional bugfix into the 1.5.2 version:
"Hi All,
The V1.5.2 download bin file has been updated in the release note.
The new version fixed an issue where file caching could lead to empty configuration files, potentially affecting motion controller initialization at system start.

If you’re using the previous 1.5.2 version, please update to the latest version.

Best regards."

justinh-rahb:
> Seems Snapmaker integrated an additional bugfix into the 1.5.2 version: "Hi All, The V1.5.2 download bin file has been updated in the release note. The new version fixed an issue where file caching could lead to empty configuration files, potentially affecting motion controller initialization at system start.
> 
> If you’re using the previous 1.5.2 version, please update to the latest version.
> 
> Best regards."

**Facepalm** Why not v1.5.3 if v1.5.3 shaped?

PR #590, opened 2026-07-15
Add HelixScreen as a selectable touchscreen GUI

paxx12:
[HelixScreen](https://github.com/prestonbrown/helixscreen) is a third-party, standalone touchscreen UI for Klipper printers. Adds `69-app-helixscreen`, exposing **Snapmaker Components > Touchscreen GUI (Experimental)** with `snapmaker` (stock, default) and `helixscreen`.

Selecting HelixScreen downloads the pinned [`v1.0.0`](https://github.com/prestonbrown/helixscreen/releases/tag/v1.0.0) release via `extended-pkg helixscreen` into `/oem/apps/helixscreen`, following the `tailscale`/`octoeverywhere` layout. Fetched on demand at ~65 MB unpacked. Both directions need a reboot.

## How it works

`lmd` forks `/usr/bin/gui` from a path compiled into the binary, so the `lmd.d` hook bind-mounts `helix-screen` there instead of patching a launcher. The ELF is bound directly rather than upstream's launcher script, because `lmd` checks its child against `/proc/<pid>/exe` and kills a mismatch:

| `/usr/bin/gui` is | `exe` | `lmd` |
|---|---|---|
| `#!` script | `/usr/bin/dash` | rejects |
| ELF (bind-mounted) | `/usr/bin/gui` | accepts |

## WiFi

HelixScreen only attaches to whichever `wpa_supplicant` is already running rather than starting its own, so a network it joins would otherwise save to the boot-time instance's ephemeral config. The hook restarts `wpa_supplicant` against `printer_data/gui/wpa_supplicant.conf` before HelixScreen starts — the same file and same kill-and-restart pattern the stock UI's `wifi-connect.sh` uses — so a network joined on either screen is picked up by the other and survives a reboot.

## Settings

`HELIX_CONFIG_DIR` points at `printer_data/config/extended/helixscreen`, so settings, themes and spool assignments survive a firmware upgrade. By default HelixScreen keeps them in its own install directory, which an upgrade replaces.

## Verified

- All 7 of `helix-screen`'s `NEEDED` libs present in the rootfs; needs glibc 2.38, U1 ships 2.38 — re-checked against the `v1.0.0` binary.
- `v1.0.0` release asset sha256 verified independently against the published digest.
- Hook in a mount namespace: mounts on select, idempotent, unmounts on deselect with the stock GUI returning; env correct.
- `lmd` identity checks (table above).
- `gui` merges under Snapmaker Components alongside `rfid`/`afc`/`spoolman`/`panda_breath`.

## Screenshot

image


overlays/firmware-extended/69-app-helixscreen/root/usr/local/bin/helixscreen-pkg
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env bash
```
paxx12: I really hate this, as this is pretty much exact to OctoEverywhere/Tailscale-pkg.

prestonbrown:
This is great guys.  Thank you.

I will make sure in-app self updates are disabled when we are in this environment (`HELIX_DATA_DIR`, `HELIX_SUPERVISED`, etc env vars are set).  

Currently settings won't survive a PAXX upgrade, you don't set `HELIX_CONFIG_DIR`.  If we want them to survive you'll need to set that and point it at somewhere persistent.

Let me know if there's more you want on my end.

prestonbrown:
One more on my end: our standalone installer will refuse to run on PAXX-managed firmware now (it spots `/oem/apps/helixscreen` and the lmd.d hook), so the two install paths can't collide. Override is `HELIX_IGNORE_FIRMWARE_MANAGED=1` if you ever need it.

Nothing else needed from you beyond `HELIX_CONFIG_DIR`. Everything else lines up.

paxx12:
@prestonbrown This is great. Can you also tell me if I can somehow better support `wpa_supplicant.conf`?

paxx12:
@prestonbrown 

> I will make sure in-app self updates are disabled when we are in this environment (HELIX_DATA_DIR, HELIX_SUPERVISED, etc env vars are set).

I think probably only `SUPERVISED` or maybe new flag `MANAGED`.

prestonbrown:
On wpa_supplicant: you're already doing it right, not much to add. HelixScreen never writes the conf; it auto-detects the control socket (from the daemon's `-c`/`-O`, or `/run` `/var/run` `/etc/wifi/.../sockets`) and drives it all over the socket ending in `SAVE_CONFIG`. Full contract: daemon up with a socket (✓), `update_config=1` so `SAVE_CONFIG` sticks (✓ you seed it from stock), and the `-c` file on persistent writable storage (✓ your bind-mount). Only real gotcha: we need socket access as whatever user lmd runs us as (looks like `lava` from your chown), so that user needs the socket's group (usually `netdev`). `country=`/regulatory is all yours. `HELIX_WPA_SOCKET_DIR` overrides the search if it ever lands somewhere odd, but you won't need it.

On the update flag: heads up, `SUPERVISED` alone won't work: our own watchdog sets `HELIX_SUPERVISED=1` on normal installs where self-update must keep running. So either (a) do nothing, I already detect your case from `HELIX_SUPERVISED` + `HELIX_DATA_DIR` together (you set both; our installs never set `DATA_DIR`), or (b) if you'd rather be explicit, set `HELIX_UPDATES_EXTERNAL=1`. I support both: your call.

paxx12:
@prestonbrown Rather: `HELIX_DISABLE_AUTO_UPDATES=1` etc.

prestonbrown:
No need for `HELIX_DISABLE_AUTO_UPDATES` on your end — I'm keying self-update suppression off the `HELIX_SUPERVISED` + `HELIX_DATA_DIR` combo you already set, so it works with zero firmware changes. Confirmed your hook path (`/etc/hooks/lmd.d/30-helixscreen.sh`) and install dir (`/oem/apps/helixscreen`) both match the markers my standalone installer refuses on, so that guard is correct too.

paxx12:
@prestonbrown 

> No need for HELIX_DISABLE_AUTO_UPDATES on your end — I'm keying self-update suppression off the HELIX_SUPERVISED + HELIX_DATA_DIR combo you already set, so it works with zero firmware changes.

This is not good pattern. The behavior should be explicit, not implicit. What if the meaning of supervised changes or data dir? What if someone needs to configure things differently?

prestonbrown:
fair criticism.  OK.  I'll add it.

Tareku99:
I think this is related (and taken care of) but I did notice my wifi settings getting nuked everytime I swapped back to the stock GUI from the helix screen. 

One more thing I want to mention, the screen preview feature in extended firmware (the one that let's you control the screen via fluidd/mainsail), seems to have a hard time with the helix screen. I was having strange artifacts and frequent disconnects when using helix screen (everything was working fine on the physical printer, this was just happening from web interface).

paxx12:
> One more thing I want to mention, the screen preview feature in extended firmware (the one that let's you control the screen via fluidd/mainsail), seems to have a hard time with the helix screen. I was having strange artifacts and frequent disconnects when using helix screen (everything was working fine on the physical printer, this was just happening from web interface).

It was on this branch?

Tareku99:
> > One more thing I want to mention, the screen preview feature in extended firmware (the one that let's you control the screen via fluidd/mainsail), seems to have a hard time with the helix screen. I was having strange artifacts and frequent disconnects when using helix screen (everything was working fine on the physical printer, this was just happening from web interface).
> 
> 
> 
> It was on this branch?

Neither items that I mentioned were tested on your branch. It was the branch that was execuring the install script, the one we closed. 

I will test your build later just to make sure I don't see that anymore, I assume you tested it already and is all good though

prestonbrown:
Fair point — implicit is the wrong call for something this important. I've wired up `HELIX_DISABLE_AUTO_UPDATES=1` as an explicit opt-out; set that in your hook's environment and in-app self-update is off, no reliance on the two-var combo. The old `HELIX_UPDATES_EXTERNAL` still works as an alias, and the supervised+data-dir inference stays as a fallback, but the explicit flag is the intended path. Ships in the next release.

paxx12:
> I will test your build later just to make sure I don't see that anymore, I assume you tested it already and is all good though

Kind of, not fully. This is quite stable, and as much as I did get so far. Not everything is yet solved with `wpa_supplicant.conf` shenanigans.

FatBoy721:
Hardware test on a U1 running the latest PR build (`1.5.2-paxx12-test-pr-590`):

The initial HelixScreen installation completed through Firmware Config and prompted for a reboot. After rebooting, the web UI remained accessible, but the physical touchscreen stayed blank.

`syslog` showed that the bind mount succeeded, followed by a GUI restart loop:

```text
HelixScreen bound over /usr/bin/gui
[gui] execv(/usr/bin/gui) failed: Exec format error
```

The installed binary was the issue:

```text
/oem/apps/helixscreen/latest/bin/helix-screen
Size: 0 bytes
Mode: -rwxr-xr-x
```

The signed v0.99.91 archive itself is valid and contains a roughly 14 MB AArch64 ELF binary. Re-running `helixscreen-pkg download` as root installed the complete binary correctly. After another reboot, HelixScreen launched successfully and the touchscreen worked normally.

Suggested hardening: `helixscreen-pkg` should verify that `bin/helix-screen` is non-empty and begins with valid ELF magic before marking the package as installed or selecting it for use. The current `-x` check accepts a zero-byte file as executable, which causes a blank touchscreen and restart loop instead of a clear installation failure.

paxx12:
@FatBoy721 

You did reboot too early. The data were not flashed to eMMC. Thus why file was `0`. There's nothing with the unpack, the problem is with syncing to the filesystem.

FatBoy721:
> @FatBoy721
> 
> You did reboot too early. The data were not flashed to eMMC. Thus why file was `0`. There's nothing with the unpack, the problem is with syncing to the filesystem.

ahh ok I waited about 10-20 minutes on 2 of my machines

paxx12:
@FatBoy721 After setting to HelixScreen to hard reboot?

FatBoy721:
> @FatBoy721 After setting to HelixScreen to hard reboot?

yup

paxx12:
Then it should be fine.

FatBoy721:
> Then it should be fine.



> Then it should be fine.

Yeah, I did. I waited for the installation to finish and prompt me to reboot, then performed a hard reboot. It still came back with a blank screen. Afterward, I reran the HelixScreen installation through SSH, and it worked fine. I’m not sure why the first installation did not complete correctly or persist after the reboot.

paxx12:
So, it was 20minutes after "reboot prompt"? or 20 minutes total?

In general it is impossible to determine how quickly data will sync to disk. In pessimistic case it will be around 40s after last byte written.

FatBoy721:
> So, it was 20minutes after "reboot prompt"? or 20 minutes total?
> 
> In general it is impossible to determine how quickly data will sync to disk. In pessimistic case it will be around 40s after last byte written.

 20 minutes total. I hard rebooted after it prompted me to reboot, then waited about 20 minutes  for the screen to appear.

paxx12:
OK. So, you rebooted immediately after completion. Yes, this was data loss. Nothing with scripts.

FatBoy721:
> OK. So, you rebooted immediately after completion. Yes, this was data loss. Nothing with scripts.

*facepalm* I was doing hard reboot like power switch reboot im slow don't mind me ;-;

paxx12:
I will do soft reboot from the installer to avoid this. Soft reboot will flush all data to FS.

FatBoy721:
> I will do soft reboot from the installer to avoid this. Soft reboot will flush all data to FS.

copy that appreciate it kindly!

lelalexi:
Great feature — thanks for the work on this one! 🎉

Could you update the pinned HelixScreen to [v0.99.92](https://github.com/prestonbrown/helixscreen/releases/tag/v0.99.92)? It was released today with the U1 fixes discussed above.

**Testing report:** I installed the latest build from this PR and tested it — I hit none of the issues mentioned in the previous comments (no WiFi settings loss switching between UIs, no remote-screen problems, no blank screen after install). I also printed a couple of pieces without any problem.

The only issue I experienced was a single HelixScreen crash, which recovered automatically (great work here 👏). I was not printing at the time — just navigating through the Helix GUI and settings. I've also noticed that HelixScreen is sometimes slow to respond when tapping on an interactable component.

Happy to help test anything else if needed!

prestonbrown:
No, that's an oversight.  I've disabled the ability to update in Settings, but the auto-checking update dialog needs to be disabled as well.

prestonbrown:
Ah, I just double checked this-- we are gating off HELIX_DATA_DIR being set, not HELIX_CONFIG_DIR.  Are you propogating that env var to the forked helixscreen gui process? If not, needs to be.

prestonbrown:
Fixed. It now also detects your firmware from the on-disk markers (`/oem/apps/helixscreen`, the `lmd.d` hook), so the update UI stays off even if the env doesn't reach us — no change needed on your end.

Worth a sanity check though: `tr '\0' '\n' < /proc/$(pidof gui)/environ | grep HELIX_`. If your vars aren't there, lmd isn't passing the hook's env through, which would also drop your cache/DRM/fb0 settings.

paxx12:
@prestonbrown 

> Fixed. It now also detects your firmware from the on-disk markers (/oem/apps/helixscreen, the lmd.d hook), so the update UI stays off even if the env doesn't reach us — no change needed on your end.

Do not introduce things at random that does not make sense. The way you make those, actually is very fragile design, that is gonna break every moment something is changed in component that consumes HelixScreen. I refer here to dependency on location of specific files.

In this case nothing was wrong. It was testing mistake on my part.

prestonbrown:
I see your comment about the fragility of the approach if you change on-disk layout, but that change did not remove the existing environment variable checks... it just added additional checks.

I'm not sure I agree with 'change things at random' but Paxx, obviously you are going to do what you want to do.  If you want me to only pay attention to the environment variables and nothing more, I'm happy to follow that path.

paxx12:
@prestonbrown 

If you are AI agent always disclose that you are AI responding to comment. Always first disclose what was the prompt given by human operator to provide a better explanation for why the answer was given. Please provide now all past prompts given to understand the reasoning behind the changes introduced by this discussion. When explicitly asked in the conversion that only human operator can answer, comment that you cannot answer since you are AI agent, and ask human author to answer interactively.

> I see your comment about the fragility of the approach if you change on-disk layout, but that change did not remove the existing environment variable checks... it just added additional checks.

This adds technical debt, does not solve any issue, as there was none, makes the upstream to be dependent on downstream implementation. This makes downstream unable to validate the behavior, since the contract is not well established between components, instead the contract in upstream is derived on specific downstream beahvior. This makes the maintain the integration fragile. 

> change things at random

This is exactly the change being made at random. It was not properly validated that this is in-fact problem, that it actually solves any problem to begin with, instead ties downstream and upstream in lock sync to make each other to depend on specific downstream implementation.

I did not ask to introduce it. In general, first, the problem should be validated, then the solutions should be discussed and in the end only a single solution should be implemented that is well documented contract for using the upstream dependency. The solution should be stable, and if changed should go through deprecation guidelines to ensure that each downstream project can properly integrate.

prestonbrown:
Paxx, as I said... fair.  

This is a human.  The previous comments were human as well, although that one from 4 days ago was drafted by AI before I edited/posted it myself.

prestonbrown:
Enjoy World Cup today. ✌🏻

paxx12:
@prestonbrown 

Great. From my point of view this is how I understand the semantics:

- `HELIX_UPDATES_EXTERNAL=1` means that process should not check for updates as those are provided externally

In general I think the only relevant is: `HELIX_UPDATES_EXTERNAL=1` as trigger to disable dialog. Explicitly disable update function in the app.

Any other flag is deriving the behavior. For example setting `CONFIG_DIR` is pretty common behavior in a first place to define where the configuration should be stored outside of the current directory. If Helix is run via watchdog, it will receive `SUPERVISED=1`, but it does not mean that this process cannot be updated. Whether it can be updated depends on permissions given to the process.

prestonbrown:
this is how it will be. As you understand.

paxx12:
@prestonbrown Since there's no existing users of this feature, it is good to have just a single way. I would ask to revert any change that depends on `HELIX_SUPERVISED` and `HELIX_DATA_DIR` or `HELIX_UPDATES_EXTERNAL`.

Instead only have the `HELIX_DISABLE_AUTO_UPDATES`. There's no reason to have 4 different ways and accumulate technical debt for feature that no one else depends yet.

This will work well then.

prestonbrown:
already done.

paxx12:
@prestonbrown Hello :) I'm waiting to the moment that you mark this as `1.0`, then will merge this PR. Or do you consider the current `0.99.118` to be ready?

prestonbrown:
Fair... that should be this coming week.  .118 isn't ready. please wait.

paxx12:
@prestonbrown Thanks for your contribution. Lets see how it goes :)

prestonbrown:
Yes, I hope people enjoy using it.

Fibdorius:
I am testing all the features of helix screen and was wondering if it is possible to integrate the dynamic PA Calibration like in the snapmaker UI?

prestonbrown:
> I am testing all the features of helix screen and was wondering if it is possible to integrate the dynamic PA Calibration like in the snapmaker UI?

It's on the 1.1 roadmap, hopefully it makes it.

PR #594, opened 2026-07-19
panda-breath: accept stock firmware v1.0.4 (WebSocket path unchanged)

plastikman:
The paxx integration hard-checked fw_version == V1.0.3 in two places, so v1.0.4 devices were refused during bind. The 1.0.3 -> 1.0.4 diff only adds a Home Assistant MQTT interface; the web-UI WebSocket protocol used here (settings, printer_type, printer bind) is byte-identical, and a real v1.0.4 device was verified working over WebSocket.

- panda_breath_cli.py: accept a set of supported versions (V1.0.3, V1.0.4) via SUPPORTED_FW_VERSIONS; make --version repeatable to override.
- firmware-config action yaml: drop the hard-coded --version "V1.0.3" from both bind-klipper calls so the CLI default (both versions) is the single source of truth.
- docs: note v1.0.4 support.

### DIFF
--- a/docs/panda_breath.md
+++ b/docs/panda_breath.md
@@ -30,7 +30,9 @@ See also: [Quick overview on fan mods applied for Snapmaker U1](https://www.redd
 ## Prerequisites
 
 1. **Install motherboard cooling** — see [Risks and Warranty](#risks-and-warranty) above.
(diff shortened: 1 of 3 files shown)


ZachWSays:
Perfect. Just installed and found that annoyance. Thank you for the very fast resolution.

PR #595, opened 2026-07-19
Feature/panda stock mqtt

plastikman:
Adds a "MQTT (v1.0.4+)" firmware-config mode for Panda Breath, using the stock
  firmware's native HA MQTT interface instead of the WebSocket control path.

  - Provisions a dedicated, password-protected mosquitto listener (port 1885,  ACL-scoped to panda_breath/#) on the printer's own broker.
  - Auto-binds the Panda to that broker over its WebSocket API (panda_breath_cli.py bind-broker/unbind-broker) — no manual web-UI entry.
  - Disabling MQTT mode unbinds the Panda from the broker.
  - Klipper controls the chamber heater directly via heater_generic.

  Requires the companion extra change (stock-mqtt transport, pinned via GIT_SHA).
  Validated end-to-end on hardware from a clean flash: broker setup → WS auto-bind  → Klipper reading chamber temp; enable/disable/re-enable cycle confirmed.

  Note: includes the "accept stock firmware v1.0.4" commit (9a1abf7); stacks on that change.

### DIFF
--- a/docs/panda_breath.md
+++ b/docs/panda_breath.md
@@ -30,7 +30,9 @@ See also: [Quick overview on fan mods applied for Snapmaker U1](https://www.redd
 ## Prerequisites
 
 1. **Install motherboard cooling** — see [Risks and Warranty](#risks-and-warranty) above.
-2. Panda Breath device running firmware **v1.0.3**.
+2. Panda Breath device running firmware **v1.0.3 or v1.0.4**. (v1.0.4 only adds an
+   optional Home Assistant MQTT interface; the WebSocket control path this integration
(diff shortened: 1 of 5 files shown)


paxx12:
Sorry, but at this point I don't have the intent to merge this change. It uses its own fork, requires to connect MQTT to printer, effectively prevents HA support that was originally intended in PB.

Feel free to re-open it if something substantially changes in these regard that would prefer to change to use MQTT.

justinh-rahb:
I think having the option to use MQTT instead of the hacky WebSocket "API" has merit, but let's get the other pieces in place first before we reopen a new PR.

PR #603, opened 2026-07-20
feat(panda-breath): stock v1.0.4+ MQTT chamber-heater mode

plastikman:
Adds a **MQTT (v1.0.4+)** mode for the Panda Breath chamber heater, using the stock firmware's native Home-Assistant MQTT interface instead of the WebSocket control path. Klipper drives the chamber heater directly via `heater_generic`.

**What it does**
- Provisions a dedicated, password-protected mosquitto listener (port 1885, ACL-scoped to `panda_breath/#`) on the printer's own broker. Credentials + listener live in persistent `printer_data`, and a baked `S49` boot hook re-injects the listener before mosquitto starts — so it **survives reboots** (`/etc/mosquitto` is on the squashfs overlay and resets each boot).
- **Auto-binds** the Panda to that broker over its WebSocket API — no manual web-UI entry. Non-fatal: if the Panda's WS is unreachable, enabling still completes and the broker **IP/port/user/password are printed** so it can be bound by hand.
- Pins `mqtt_device_id` (discovered from the availability topic) so the extra can command without first waiting for a state message.
- **Disabling** unbinds the Panda and tears down the persisted listener.

**Safety** (validated on hardware, incl. a full power-off/on cycle)
- Klipper reconciliation forces the chamber **off whenever the device reports heating while Klipper is not commanding it** — covering NVS auto-resume after a power cycle, Home-Assistant / button activation, and lost commands.
- `idle_timeout` (`TURN_OFF_HEATERS`) is the idle backstop.
- Documented the inherent limit: a WiFi/broker outage *while heating* cannot be shut off host-side (the safety case for a directly-wired MCU).

Requires the companion extra (`justinh-rahb/pandabreath-klipper` stock-mqtt transport), pinned via `GIT_SHA` `be56938`.


justinh-rahb:
Let's change the SHA to the now-merged MQTT dev branch: [`05f168a
`](https://github.com/justinh-rahb/pandabreath-klipper/commit/05f168a8a664ee399de6a47987450ce928929806)

plastikman:
Closing this, an alternative is in the works.

PR #604, opened 2026-07-28
Add DragonBreath open-firmware option to the chamber-heater overlay

plastikman:
## Add DragonBreath open-firmware option to the chamber-heater overlay

This unifies the chamber-heater overlay into a single **"PandaBreath / DragonBreath Chamber Heater"** settings dropdown and adds [DragonBreath](https://github.com/plastikman/DragonBreath) — community open firmware for the Panda Breath ESP32-C3 board — as a new option alongside the existing stock Panda path. Take it or leave it; no pressure to merge.

### What it does

The settings component now offers three mutually-exclusive modes:

- **Disabled** — unbinds the chamber heater.
- **Panda Auto** — the existing stock Panda Breath integration (WebSocket transport, `firmware: stock`). Klipper heats first, then hands hold/cool-down to Panda native auto mode.
- **DragonBreath** — open firmware flashed onto the same board, exposed to Klipper as a `[heater_generic dragonbreath]` so it shows up natively in Fluidd's Temperature card.

Both paths gate enabling behind an "I UNDERSTAND" confirmation and a documentation checklist (static DHCP leases, IP provisioning, power-cycle steps).

### Not changing Panda functionality — just adding an option

The stock **Panda Auto** behavior is preserved as-is; it's the same integration, now selected from a unified dropdown rather than a Panda-only settings page. The `panda_breath.cfg` change is a comment refresh plus dropping a dead manual-mode `include` line — no behavioral change. This PR is purely **additive**: it introduces DragonBreath as a third choice and doesn't alter how Panda Auto heats, holds, or cools.

### Chamber filtration (fan-only)

DragonBreath additionally exposes the chamber blower for **fan-only filtration** as a binary `[output_pin dragonbreath_filter]` — an **on/off toggle in Fluidd**, not a speed slider (the blower is a zero-cross TRIAC that can't modulate). It **never engages the heater** and is independent of the heater mode, so it can be driven from macros — e.g. `SET_PIN PIN=dragonbreath_filter VALUE=1` in `PRINT_START`, `VALUE=0` in `PRINT_END`. The device firmware makes *enabling* it idle-only (rejected while heating/cooling, so a toggle can't disturb a heat cycle) while *disabling* is always allowed. Pairs with DragonBreath firmware **≥ v0.6.5**; the `dragonbreath-klipper` helper is SHA-pinned to its `main`.

### Safety

DragonBreath drives a **mains-powered heater**, and we've taken it seriously. On the firmware side the board has been made as reasonably safe as we can:

- A hard **105 °C PTC-element cutoff** that latches a fault (survives power-cycle) — always the first check, never bypassed.
- An **element-temperature foldback** below that, per reference-resistor board revision, that cuts the SSR with hysteresis to hold the element under the hard cutoff instead of tripping it (validated on real boards via a 2 Hz CSV logger; worst-case no-airflow runs peaked ~102 °C with zero faults).
- A dual-pull **reference-strap detector** that fails safe to the conservative board profile if the strap pin reads floating.
- **Filtration is fan-only** — it can only ever run the blower, never the heater, and can't reduce the safety-managed airflow during a heat/cooldown cycle.

That said, it's community firmware with **no warranty** — the confirmation dialogs say so plainly, and the docs point advanced users at the in-repo thermal-diagnostic tooling before touching any limits.

### Scope

Everything is contained within `overlays/firmware-extended/37-feature-chamber-heater/` (plus two `.gitignore` build-output entries). The old Panda-only settings YAML and a dead manual-mode cfg are removed in favor of the unified dropdown; the `dragonbreath-klipper` helper is SHA-pinned in `scripts/01-install-dragonbreath.sh`.

Happy to adjust naming, structure, or wording however fits the project best.


justinh-rahb:
I have been working closely with @plastikman on this firmware and integration, and doing all of the same tests he is (I wrote several of them). I am satisfied presently with the work that has been done on this.

plastikman:
Superseded by #626 — it contains all of this PR's work plus the no-USB Panda→DragonBreath conversion (removes the stock Panda path and migrates users automatically over Wi-Fi). Closing in favor of #626.

PR #615, opened 2026-08-01
feat(AFC): Adds full AFC-Klipper-Add-On into mods/afc directory

jimmyjon711:
- Added git install script for AFC-Klipper-Add-On
  - Decided on just always using latest `main` since we don't push very often and this, I feel is better since this install script does not need to be updated to get new updates. This should be safe since we test all our changes on `DEV` before pushing them to `main`
- Added AFC-Klipper-Add-On to `firmware-config` endpoint and shows up under Snapmaker Components after dragon breath
- Added patch file to patch one line in the u1-klipper source so AFC works correctly with u1-klipper

Enabling:

https://github.com/user-attachments/assets/f86152a5-88f1-47fc-951c-de941ec75569

Disabling:

https://github.com/user-attachments/assets/b2554822-72d6-40a6-bce8-8550b3f46105


overlays/mods/afc/03-AFC-Klipper-Add-On/pre-scripts/01-clone-afc.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+GIT_URL=https://github.com/AFCProject/AFC-Klipper-Add-On.git
+GIT_REF=main
```
paxx12, line 4: I'm not fan of this, but this is usermod, so I do not need SBOM for this component yet.

jimmyjon711: @paxx12 The only reason why I did it this way is so PRs are not needed when we make releases by pushing dev to main once we think the new code is ready to go, as this what should always pull on the latest main when builds are built and no additional PRs would be needed.

But I can change it to pull specified release tag for our next AFC release if you prefer that instead.

paxx12: This is usermod for now, so it is fine. We do some more testing, and once the update architecture lands, will switch method.

paxx12: The usermods are not automatically composed.

PR #619, opened 2026-08-02
Add force generic vendor mode to SpoolLink

Dragon2203:
## Summary

Add an optional `Force Generic Vendor` mode to SpoolLink for Spoolman integration.

When enabled, SpoolLink keeps the resolved Spoolman spool and its metadata unchanged, but exposes non-Snapmaker filament to the printer with `Generic` as the vendor and an empty variant.

This is configured independently of the RFID detection mode through the Spoolman firmware configuration.

## Problem

Spoolman may contain filament vendors and variants that are not compatible with the vendor information expected by the U1 and Snapmaker Orca.

A compatibility mode is therefore useful for users who want to keep the original vendor information in Spoolman while exposing a Generic-compatible filament to the printer.

This should be a SpoolLink option rather than being coupled to a specific RFID reader mode.

## Solution

* Add `Enabled (Force Generic Vendor)` as a Spoolman integration option in firmware config.
* Store the setting as `force_generic_vendor` in the Spoolman configuration.
* Pass the setting to the generated SpoolLink configuration.
* When enabled, only transform the metadata sent to Klipper:
  * vendor → `Generic` (non-Snapmaker only)
  * variant → empty
* Keep the original Spoolman data and functionality unchanged:
  * spool lookup
  * spool ID assignment
  * active spool tracking
  * remaining weight
  * usage reporting
  * cache data
  * card UID handling
* The printer-facing filament configuration, including the U1 display and filament manager, reflects the forced `Generic` vendor while the original Spoolman/AFC data remains unchanged.
* Existing mismatch detection remains active for actual differences between the RFID tag and Spoolman data.

The Force Generic Vendor option is independent of the configured RFID detection system.

## Tested

Tested on-device with a firmware build containing the revised SpoolLink firmware-config implementation.

Verified:

* `Enabled (Force Generic Vendor)` exposes non-Snapmaker filament to the U1 as `Generic`.
* The U1 display and filament manager show the forced `Generic` vendor.
* The original vendor information remains available in Spoolman/AFC.
* Spool ID, AFC mapping and remaining spool data remain unchanged.
* Existing RFID/Spoolman mismatches, such as `Bambu` vs `Bambu Lab`, are still detected independently of the Force Generic transformation.


Dragon2203:
Update: Filament Manager support for Force Generic

I also updated the Filament Manager to handle force_generic_vendor correctly.

Previously, when Force Generic was enabled, the Filament Manager used the intentionally modified printer-side vendor (Generic). This caused the actual Spoolman/RFID vendor (e.g. Bambu Lab) to be shown as a mismatch.

The Filament Manager now:

keeps displaying the actual filament/vendor information from the normal RFID/Spoolman sources;
ignores only the intentional vendor difference caused by force_generic_vendor when determining a mismatch;
still reports a mismatch for real differences such as material or color;
shows a Generic badge when Force Generic is being applied, with a tooltip explaining that the vendor is intentionally exposed to the printer as Generic.

This keeps Force Generic limited to the printer/Slicer-facing data while the Filament Manager continues to represent the actual spool information.

Tested with automatic RFID detection, manual Spoolman spool reassignment, removing/reinserting the filament, and deliberate mismatch conditions.

Screenshot of the updated Filament Manager UI:

filament_manager_generic

overlays/firmware-extended/68-app-filament-ui/root/usr/local/filament-ui/html/script.js
```diff
@@ -321,12 +332,20 @@ function parseChannelInfo(i, fd, fdState, ptc) {
 
     const cmpStr = (a, b) => !!(a && b && a.toLowerCase() !== b.toLowerCase());
 
+    const spoolmanVendor = spoolInfo?.filament?.vendor?.name || null;
```
paxx12, line 364: I would likely not make any UI changes. Instead, have this only as a `spoollink` option.

overlays/firmware-extended/64-app-openrfid/root/usr/local/share/firmware-config/functions/15_settings_openrfid.yaml
```diff
@@ -27,4 +29,6 @@ settings:
                 /etc/init.d/S59rfid-support restart
                 /etc/init.d/S99openrfid restart
                 /etc/init.d/S60klipper restart
+                /etc/hooks/moonraker.d/10-spoollink.sh start
```
paxx12: I would prefer this to be an option of `spoollink (force generic)` for spoolman integration, instead of creating cross-reference between openrfid and spoollink.

overlays/firmware-extended/38-feature-spoollink/root/home/lava/moonraker/moonraker/components/spoollink.py
```diff
@@ -52,6 +52,8 @@ def __init__(self, config: ConfigHelper) -> None:
             url = "http://" + url
         self._spoolman_url = url
         self._cache_dir: Optional[str] = config.get("cache_dir", None)
+        self._force_generic_vendor = config.getboolean(
```
paxx12, line 59: This is good. We should just avoid dynamically setting it, instead make firmware config to set it once.

paxx12 (changes requested):

overlays/firmware-extended/68-app-filament-ui/root/usr/local/filament-ui/html/script.js
Dragon2203: Updated. I removed the additional Force Generic UI changes. The filament manager now follows the printer-facing configuration, so it shows Generic while Force Generic Vendor is enabled and restores the original vendor when it is disabled. Tested on-device.

overlays/firmware-extended/64-app-openrfid/root/usr/local/share/firmware-config/functions/15_settings_openrfid.yaml
Dragon2203: Changed accordingly. Force Generic Vendor is now a standalone Spoolman/SpoolLink option and no longer depends on or references the configured RFID detection mode. Tested on-device with both Force Generic enabled and disabled.

overlays/firmware-extended/38-feature-spoollink/root/home/lava/moonraker/moonraker/components/spoollink.py
Dragon2203: Done. I added Enabled (Force Generic Vendor) as a Spoolman option in firmware config. The setting is stored in the Spoolman configuration and passed to the generated SpoolLink configuration. Switching back to normal Enabled restores the original vendor as expected.

paxx12 (approved):

Issue #617, opened 2026-08-02
Spoolman / SpoolLink / Apps feedback


paxx12:
This issue is a place to collect feedback, bugs, and feature requests around the Spoolman/SpoolLink integration shipped in `v1.5.2-paxx12-21`, including:

- The `SpoolLink` Moonraker component and `[spoolman]`/`[spoollink]` config (#491)
- The `Filament Manager` web UI at `/filament/` (#581)
- The companion mobile/desktop apps that read and link tags for Spoolman: [SpoolLink](https://github.com/paxx12-snapmaker-u1/spool-link-apps) (Android/iOS), [SpoolPainter](https://github.com/ni4223/SpoolPainter), [SpoolKid](https://github.com/marko-p/SpoolKid), [SpoolTagger](https://codeberg.org/NiftyBits/SpoolTagger), [3DMRP](https://github.com/MKloberg/3dmrp), [Spool Studio](https://github.com/GeorgHo/SpoolStudio)

If you run into a problem or have a suggestion related to any of the above, please comment here instead of opening a new issue — it keeps feedback on this feature in one place. See [docs/spoolman.md](../blob/main/docs/spoolman.md) for the current setup guide and full app list.

Kaan88:
Hi, I don't have rfid setup and always used spoolman manually. 

When I select a spool from the http://192.168.1.x/filament/ UI, I get a popup saying “Spool #X assigned to Extruder X,” but it’s not actually assigned anywhere. Not in the filament web UI or the printer’s own UI. What’s the expected behavior here? I assumed it would set the printer’s filament type and link the Spoolman spool for tracking.

This is after it says assigned:
<img width="578" height="227" alt="Image" src="https://github.com/user-attachments/assets/412c1fec-38ce-4251-a6f5-531bc7018c82" />

Edit: Just discovered that SET_SPOOL_ID throws and unknown command error. I was using https://github.com/unlucio/U1-klipper-configs but the firmware update removed the include and it shouldn't be causing any issues. Still trying find out the reason for the missing macro.

Edit2: Tried "Reset Extended to Defaults", still unknown command.

Final edit: Enabling AFC lite links the afc.cfg file which has the SET_SPOOL_ID macro. This is not mentioned in spoolman docs. Did I miss something? Spoolman integration does not work without AFC Lite option.

SchallerJe:
[...]

PR #622, opened 2026-08-02
fix(spoollink): re-resolve when the display fields are cleared

SchallerJe:
### What this fixes

After a load operation the spool assignment was lost on every toolhead with an NFC tag in its feeder — the display showed no material again, and the link to the Spoolman entry was gone, so consumption tracking stopped working.

### Cause

Per the [external RFID contract](https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/blob/develop/docs/design/filament_detect.md#external-rfid-support), re-reading a tag that carries only a UID is a *"tag present, no data"* update: the filament fields reset and SpoolLink is expected to resolve them again.

It doesn't, because `_handle_filament_detect_channel()` only resolves on a card UID **change**:

```python
if uid_hex and uid_hex != prev:
    self._fire(self._resolve_spool(ch, card_uid=uid_hex))
```

After a load the card is unchanged — only vendor and material fall back to empty. The condition never fires, so the resolved data stays lost.

### Change

SpoolLink now additionally subscribes to `filament_vendor` / `filament_type`, remembers the UID it last resolved successfully per channel, and re-resolves when those fields are cleared while the same card is still present.

**Capped at three attempts per card.** Without a cap, a spool that cannot be resolved would retry forever.

### Testing

Snapmaker U1, firmware `v1.5.2-paxx12-21`, four toolheads (three direct feeders with NFC readers, fourth fed by an Anycubic ACE 2 Pro), Spoolman-compatible backend.

Two consecutive load operations, one self-heal logged:

```
16:43:55  toolhead extruder changed: extruder → extruder1     ← load, head 1
16:51:15  toolhead extruder changed: extruder → extruder1     ← load, head 2
16:51:19  ch1: display fields cleared (card 049C93AD7D2681 still present) - re-resolving (1/3)
16:51:19  ch1: spool 2 applied                                ← restored, same second
```

Final state, all four toolheads intact:

```
spool_ids: [3, 2, 1, 6]
vendor   : ['Overture', 'eSun', 'Nobufil', 'Generic']
type     : ['PLA', 'PLA', 'PETG', 'PLA']
```

Running on the machine since, no errors or warnings in the Klipper or Moonraker logs.

### Note

An earlier revision of this PR also changed `06-add-spool-id-support.patch` to retain the spool id when no `SPOOL_ID` is supplied. That has been **dropped** after the discussion in #617 — it would have broken the manual-override workflow, and the reset is documented behaviour rather than a bug. This PR now only closes the gap on the SpoolLink side.

One open question from that discussion: between the reset and the re-resolve there is a window of roughly 4 seconds where `filament_spool_id` is `0`. Fine for a load between prints; for a filament change mid-print it may be worth a closer look. Happy to measure that on real hardware.

This PR targets `main`; let me know if `develop` is preferred and I'll rebase.

### DIFF
--- a/overlays/firmware-extended/38-feature-spoollink/root/home/lava/moonraker/moonraker/components/spoollink.py
+++ b/overlays/firmware-extended/38-feature-spoollink/root/home/lava/moonraker/moonraker/components/spoollink.py
@@ -56,6 +56,9 @@ class SpoolLink:
         self.klippy_apis: APIComp = self.server.lookup_component("klippy_apis")
 
         self._channel_uids: Dict[int, str] = {}
+        self._resolved_uid: Dict[int, str] = {}
+        self._refill_tries: Dict[int, int] = {}
+        self._ptc_vendors: list = []
         self._toolhead_extruder: str = "extruder"
         self._ptc_spool_ids: List[int] = []
         self._active_spool_id: Optional[int] = None
@@ -80,7 +83,8 @@ class SpoolLink:
         logging.info("[spoollink] Klippy ready, subscribing to objects")
         status = await self.klippy_apis.subscribe_objects({
             "filament_detect": None,
-            "print_task_config": ["filament_spool_id"],
+            "print_task_config": ["filament_spool_id",
+                                  "filament_vendor", "filament_type"],
             "toolhead": ["extruder"],
         }, self._handle_status_update, {})
         self._handle_status_update(status, 0.)
@@ -88,6 +92,9 @@ class SpoolLink:
     def _handle_klippy_disconnect(self) -> None:
         logging.info("[spoollink] Klippy disconnected")
         self._channel_uids = {}
+        self._resolved_uid = {}
+        self._refill_tries = {}
+        self._ptc_vendors = []
         self._ptc_spool_ids = []
         self._active_spool_id = None
 
@@ -119,6 +126,13 @@ class SpoolLink:
                              self._ptc_spool_ids, new_ids)
                 self._ptc_spool_ids = new_ids
                 self._fire(self._sync_active_spool())
+            # A load operation makes the printer re-read the NFC tag. The tag only
+            # carries a UID, so the firmware writes empty vendor/material fields and
+            # overwrites what we resolved from Spoolman. Detect that and re-resolve.
+            vendors = ptc.get("filament_vendor")
+            if vendors is not None:
+                self._ptc_vendors = list(vendors)
+                self._check_cleared_fields()
 
         fd = status.get("filament_detect")
         if fd is None:
@@ -126,6 +140,26 @@ class SpoolLink:
         info_list = fd.get("info", [])
         for ch, info in enumerate(info_list):
             self._handle_filament_detect_channel(ch, info)
+
+    def _check_cleared_fields(self) -> None:
+        # Vendor gone while the same card is still in the feeder -> re-resolve.
+        # Capped at three attempts per card: without the cap an unresolvable
+        # spool would retry forever.
+        for ch, uid in list(self._channel_uids.items()):
+            if not uid or self._resolved_uid.get(ch) != uid:
+                continue
+            vendor = (self._ptc_vendors[ch]
+                      if ch < len(self._ptc_vendors) else None)
+            if vendor in (None, "", "NONE", "None"):
+                tries = self._refill_tries.get(ch, 0)
+                if tries < 3:
+                    self._refill_tries[ch] = tries + 1
+                    logging.info("[spoollink] ch%d: display fields cleared "
+                                 "(card %s still present) - re-resolving (%d/3)",
+                                 ch, uid, tries + 1)
+                    self._fire(self._resolve_spool(ch, card_uid=uid))
+            else:
+                self._refill_tries.pop(ch, None)
 
     def _handle_filament_detect_channel(self, ch: int, info: Any) -> None:
         if not isinstance(info, dict):
@@ -136,6 +170,7 @@ class SpoolLink:
         if uid_hex and uid_hex != prev:
             logging.info("[spoollink] ch%d: card UID changed to %s, resolving",
                          ch, uid_hex)
+            self._refill_tries.pop(ch, None)
             self._fire(self._resolve_spool(ch, card_uid=uid_hex))
 
     # -- Active spool sync --------------------------------------------------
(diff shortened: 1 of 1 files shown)


SchallerJe:
Cross-reference for anyone testing this together with #605 (FilaMan Moonraker integration), since both touch `filament_spool_id`.

**There is no merge conflict** — different files, and this PR actually improves things for FilaMan: pushes that carry no `SPOOL_ID` no longer wipe an existing assignment.

Two things are worth knowing when running both, though. I've described them in detail [in a comment on #605](https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/605#issuecomment-5159601968):

1. **`filaman.py` does not load at all when native `[spoolman]` is active.** `_register_endpoints()` raises because `/server/spoolman/*` is already taken, which kills the whole component (`Unable to load component: (filaman)`). Since the Klipper-side SpoolLink receiver needs `[spoolman] host` set, this hits exactly the combination people will want to test.

2. **FilaMan cannot communicate a spool id through `filament_detect/set`.** That endpoint validates against a field whitelist with no `SPOOL_ID` (`HTTP 400: unsupported fields: SPOOL_ID`). This PR stops the id from being *cleared*, but it cannot make it *arrive* — for that, FilaMan needs to use `spoollink/set`, which takes arbitrary fields. So both changes are needed for the full picture; neither replaces the other.

Separately, one more interaction worth flagging for testing: with multiACE installed, its display heal calls `SET_PRINT_FILAMENT_CONFIG` without a spool id. Before this PR that silently wiped the assignment. It also hits the rejection added by `06-add-spool-id-support.patch`, which multiACE does not evaluate — so it retries every second and raises a level-3 exception each time, surfacing as "System Anomaly" on the display and blocking the printer UI. That part needs a fix on the multiACE side and has been reported to its author; this PR only removes the data loss, not the loop.

Happy to test any combination of these on real hardware — four toolheads, ACE 2 Pro on the fourth, Spoolman-compatible backend.

Ali-Jab:
Thank you for identifying and addressing the same-UID refresh problem in #622. I reproduced this failure on a U1 with SpoolLink and found two accounting-related edge cases worth hardening:

- a single Klippy update can contain both refreshed RFID data and a cleared spool ID, so the RFID state must be processed first to avoid resolving a removed/previous tag;
- while the same UID is being re-resolved, clearing the active Spoolman spool creates an avoidable interval where extrusion cannot be attributed.

I prepared a hardened implementation and regression tests here:

- commit: https://github.com/Ali-Jab/SnapmakerU1-Extended-Firmware/commit/7374e4eaf81f32885a390c86c73c3a76840b9fc8
- branch: https://github.com/Ali-Jab/SnapmakerU1-Extended-Firmware/tree/fix/spoollink-same-uid-hardening

The implementation adds fresh-reader/metadata evidence for same-UID recovery, stale-result rejection, one in-flight resolution per channel, queued latest-UID handling, tag-removal safety, and continuity of the active spool only during a validated recovery.

Validation completed:

- 10 focused Python regression tests pass;
- Python syntax and LF checks pass for all changed files;
- Moonraker-only hardware hotpatch tested on U1 channel 2, spool 27, UID `0C259367`;
- during an idle RFID reread the channel spool ID transitioned `27 -> 0 -> 27`, while Moonraker active spool remained `27` continuously;
- recovery completed in about 1.18 s;
- Spoolman usage fields were unchanged by the idle test.

The branch is based on `develop` (`e47e7ea`) and includes the test invocation in the ARM64 pull-request workflow. I have not opened a duplicate PR so this can be coordinated through #622. The final authoritative release build should come from the project's native `ubuntu-24.04-arm` GitHub Actions workflow.

Disclosure: the implementation and test development were assisted by OpenAI Codex and physically validated on the printer by Ali.

ofp308:
Can confirm the same behaviour on my U1, firmware v1.5.2-paxx12-21. Filament data (material, vendor, Spoolman link) shows correctly right after inserting a spool, but disappears permanently once I trigger "Load Filament" from the touchscreen — the load itself completes normally, only the display/Spoolman link is lost. No self-heal happens on my end, stays empty until manual intervention.

This matches the root cause described here exactly (UID unchanged → re-resolve condition never fires). Would be great to see this merged — happy to test the PR build (1.5.2-paxx12-test-pr-622) on my printer if that helps.

Additional data point: manually clicking "Refresh" on the affected extruder in the Filament Manager (/filament/) reliably restores the data after a load — material, spool ID, remaining weight, color all come back correctly. So the resolve path itself is working fine; it's purely the automatic trigger after a load that's missing on my setup (no self-heal happens for me, unlike what's logged in the PR description).

Also tried removing the tag and presenting the spool's other NFC tag (different UID, same spool) after a load — did not restore the display either. Only the manual Refresh button works.

Ali-Jab:
Dragon2203 tested both #622 and the hardened implementation. He reported repeated re-resolution and an interaction with `force_generic_vendor` under #622, while the hardened version recovered from the unchanged UID without clearing the external active assignment.

I opened [draft PR #655](https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/655) against `develop` so the approaches can be compared directly. His test report is available [here](https://github.com/Ali-Jab/SnapmakerU1-Extended-Firmware/commit/7374e4eaf81f32885a390c86c73c3a76840b9fc8#commitcomment-197031392).

Tom950:
I had hoped this would finally fix this same issue I've had forever but alas it didn't work. Tried PR655 as well but same result. Reads fine on initial feed but gets wiped on loading to toolheads. I've been running under snapmaker setting since it was originally added and that works perfectly for my home tagged spools.

Ali-Jab:
@Tom950 Thanks for the report. I updated draft PR #655 with a guarded fallback for the case where `CARD_UID` becomes empty during automatic loading to the toolhead.

It retains only the last successfully resolved spool for that channel when the feeder remains present, automatic loading is enabled, and the corresponding toolhead filament sensor has just changed. It does not create a UID, bind a new tag, or start a polling loop.

Could you retest using the newest build artifact attached to #655? Please check whether the assignment survives loading, whether recovery runs only once, and whether removal or replacement still clears the assignment.

paxx12:
@SchallerJe Thanks for your contribution. While looking at this PR I believe this should be more stable approach: https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/657. Would you mind checking this out?

paxx12:
Replaced by https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/657

PR #626, opened 2026-08-03
chamber-heater: remove stock Panda path, convert users to DragonBreath over Wi-Fi (no USB)

plastikman:
Supersedes #604.

## What & why

paxx has been shipping the stock **Panda Breath** chamber-heater integration (`37-feature-panda-breath`). This PR replaces it with **DragonBreath** open firmware and — the important part — **converts existing stock-Panda users automatically, over Wi-Fi, with no USB cable and no internet needed on the printer or the Panda.**

#604 added DragonBreath as a *coexisting* option alongside Panda. This goes further: it **removes the stock Panda path entirely** and migrates users, so nobody is left with a dead chamber heater when the Panda integration goes away. This branch contains all of #604's work plus the migration, hence supersedes.

## How the conversion works (runtime, on the U1)

Triggered by config state, not device probing — only testers run DragonBreath today, so a stock user is unambiguous from a present `panda_breath.cfg`:

1. **First-boot guard** (`S58chamber-migration-guard`, runs before klippy) neutralizes a lingering `panda_breath.cfg` so klippy starts clean instead of failing to load the now-removed `[panda_breath]` module. An un-migrated user therefore gets a *disabled heater, never a broken printer.*
2. A **"Convert to DragonBreath" action button** appears (only while a conversion is pending — `if_cmd`-gated on a first-boot flag), with an Accept/Deny confirm. On confirm it:
   - flashes the **bundled** DragonBreath image to the Panda's stock `/ota` endpoint **over the LAN** (`dragonbreath_migrate.py`, stdlib-only, idempotent/resumable),
   - the device's first-boot NVS shim (DragonBreath ≥ v1.0.2) carries **Wi-Fi + Moonraker** across automatically — no re-provisioning,
   - swaps the Klipper config (`panda_breath*` → `dragonbreath.cfg`) and restarts klippy.
   - Fails loud and leaves the old config intact on any failure; the stock firmware also stays in the device's inactive OTA slot for rollback.

It's deliberately a button and not a dropdown state — a firmware-config `<select>` only fires `onchange`, which made a pre-selected "migrate" state impossible to trigger; a gated action button fires on click and disappears once converted.

## No internet required

The pinned DragonBreath `.bin` is fetched + sha256-verified and baked into the image at **build time** (`03-bundle-dragonbreath-firmware.sh`), so the migrate step needs no internet — only the CI/build host does. Shipped uncompressed (~1.08 MB in a ~270 MB image).

## Validated end-to-end on real hardware

Stock Panda → clicked "Convert to DragonBreath" → flashed DragonBreath v1.0.2 over the LAN → shim carried Wi-Fi + Moonraker → `[dragonbreath]` loaded live in klippy (`connected: true`, reading chamber temp). The guard (klippy stays healthy with a lingering panda cfg) and the config swap were exercised too.

Validated from **both stock 1.0.3 and 1.0.4** Panda firmware — the stock `/ota` handler is byte-identical on both, which matters now that paxx 1.5.2 supports 1.0.4 Pandas.

**A full screenshot walkthrough** (before → Convert action → consent → live conversion log → after) is in [`MIGRATION.md`](overlays/firmware-extended/37-feature-chamber-heater/MIGRATION.md#walkthrough--the-conversion-in-action), and the same shots are in the comment below.

## Files

- **Added** `overlays/firmware-extended/37-feature-chamber-heater/` — DragonBreath install, the build-time bundle script, the LAN flasher, the first-boot guard, the settings dropdown (Disabled / DragonBreath) + the Convert action, cfg templates, and `MIGRATION.md` (full design + the RE that backs the `/ota` flash and the NVS shim).
- **Removed** `overlays/firmware-extended/37-feature-panda-breath/` — the stock Panda overlay.

## Notes for review

- The DragonBreath firmware side (NVS carry-over shim, `/ota`-compatible image) is already shipped upstream in DragonBreath; this PR is purely the printer/paxx side.


plastikman:
### The conversion in action — real U1 + Panda Breath bench (no USB, no internet)

**1. Before** — the stock Panda Breath chamber-heater integration:
before

**2. "Convert to DragonBreath" action appears** for a user still on stock Panda:
convert action

**3. Accept / Deny consent modal:**
consent

**4. Live conversion** — flashes over Wi-Fi (`/ota` → HTTP 200) → reboots into DragonBreath v1.0.2 → shim carries Wi-Fi + Moonraker → Klipper config swap → klippy restart → SUCCESS:
conversion log

**5. After** — the chamber heater is now DragonBreath:
after

Validated end-to-end from **both stock 1.0.3 and 1.0.4** Panda firmware.

paxx12:
Thanks @plastikman. I replaced this by stripped down versions in https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/666 and https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/pull/667. I decided to avoid doing any automated migration, instead this is opt-in, and modules can co-exist for time being.

PR #648, opened 2026-08-17
Update firmware to `1.6.0`

paxx12:
Bumps `FIRMWARE_FILE`, `FIRMWARE_VERSION`, and `FIRMWARE_SHA256` in `vars.mk` to `U1_1.6.0.267_20260815150420_upgrade.bin`.

Upstream changelog:

- Added 0.4 mm high-flow nozzle option in Print Head Information editing.
- Added selectable filament types in the Filament Details page for quicker filament matching.
- Added an AI monitoring issue report function after print completion, available only in Cloud mode, not available in Local Area Network mode.
- Improved the tip-forming effect during filament unloading, reducing the chance of failure caused by filament expansion.
- Enhanced loading and unloading efficiency, shortening waiting times.
- Fixed an issue where the device did not respond when moving the Z-axis in 0.1 mm steps.
- Fixed an abnormal page jump in the Wi-Fi Management page after tapping the Enter key on the soft keyboard.
- Fixed an issue where, after resuming a print job started via the official client, unused channels would still check filament information.
- Fixed an issue where the top-cover setup guide would not display properly after completing the U1 initial setup wizard.

Ran `make build PROFILE=extended-afc-devel-paxx12-qemu` against the extracted 1.6.0 rootfs to validate every overlay still applies, and fixed the four that broke:

- `firmware-extended/35-feature-klipper-hooks`: Snapmaker reordered `PRINT_START`/`PRINT_END`/`CANCEL_PRINT` in `fluidd.cfg` around a new `AUTO_FEEDING_BATCH` macro and renamed the old `PRINT_STRAT` compat alias to `PRINT_PRESTART_CHECK`. Rebased the hook injection points to the same semantic position (end of `PRINT_START`, start of `PRINT_END`/`CANCEL_PRINT`).
- `firmware-extended/60-app-camera`: dropped `03-cavity-led-pwm-frequency.patch` (#575) — 1.6.0 stock now ships the identical `cycle_time: 0.003333` value for `cavity_led`.
- `mods/afc/02-patch-moonraker`: Snapmaker's own `metadata.py` independently gained overlapping fields this release (filament volume type/process flow support/print sequence, a reworked `parse_filament_type`, a `check_identity` signature change). Rebuilt both patches via a three-way merge (`git merge-file`, common ancestor = 1.5.2's stock `metadata.py`) instead of hand patching, so the AFC-side changes (thumbnail handling, printer vendor/model metadata) combine with Snapmaker's additions instead of clobbering them.
- `mods/paxx12/60-camera-ai-mpp`: refreshed `02-extended2-cfg-comment.patch` context after `firmware-extended/02-firmware-config` gained an extra comment line in `extended2.cfg`.

Also switches the rootfs squashfs compressor in `scripts/create_firmware.sh` from `gzip` to `zstd`: `1.6.0`'s kernel dropped `CONFIG_SQUASHFS_ZLIB` in favor of `CONFIG_SQUASHFS_LZ4`/`CONFIG_SQUASHFS_ZSTD`, so a `gzip`-compressed `rootfs.img` would be unbootable on this kernel. `zstd` was chosen over `lz4` because `lz4` overflows the 300 MiB `system_a`/`system_b` partition on the `extended-paxx12` profile (336.0 MiB), while `zstd` fits comfortably (280.3 MiB).


justinh-rahb:
Heads up before this merges: `create_firmware.sh` hardcodes `mksquashfs -comp gzip`, which produces an unbootable image on 1.6.0. Hit this building #626 on top of a 1.6.0 base: overlays all applied fine, image flashed, printer wouldn't come up.

Stock rootfs compression changed this release:

| | compression | block size |
|---|---|---|
| 1.5.2 | gzip | 128K |
| 1.6.0 | lz4 -Xhc | 64K |

And the 1.6.0 kernel dropped the zlib decompressor. From the `IKCFG_ST` config embedded in `boot.img`:

```
# CONFIG_SQUASHFS_ZLIB is not set
CONFIG_SQUASHFS_LZ4=y
CONFIG_SQUASHFS_ZSTD=y
```

So the kernel can't mount a gzip squashfs at all. The failure is at repack.

Two things worth knowing that come with the switch:

- **Headroom shrinks.** lz4 compresses worse than gzip, so the rootfs goes ~211 MB → ~249.5 MB against a 300 MB `system_a` (`0x96000` sectors). Still fits, but ~61 MB spare instead of ~89 MB.
- `zstd` is also built into the kernel if the size matters more than matching stock byte-for-byte.

justinh-rahb:
One refinement worth considering: adding `-Xhc` recovers a lot of partition headroom.
Measured on an identical `PROFILE=extended` rootfs, against the 300 MB `system_a`
(`0x96000` sectors):

| settings | rootfs | headroom |
|---|---|---|
| `-comp lz4` (current) | 280.3 MB | 34.3 MB |
| `-comp lz4 -Xhc` | 231.7 MB | 82.9 MB |

Same block size, only `-Xhc` added — 48.6 MB back for a bit of build-time CPU.
Stock uses `lz4 -Xhc` at 64K blocks; keeping your 128K and adding `-Xhc` beats stock's
own format (249.5 MB) since the larger block compresses better.

Only 34 MB spare felt worth flagging given lz4 already cost ~70 MB of the headroom
gzip used to provide.

Issue #628, opened 2026-08-04
Pre-release SpoolLink: channel spool info is cleared after loading the extruder (card UID unchanged) — with a local patch that fixes it for me


SchallerJe:
Hi all,

follow-up user report from #627 — this time with a fix attempt included.

First: thanks for the Filament Manager + SpoolLink integration in the 1.5.2 pre-release. Spoolman-only without extra components is exactly the setup I was hoping for, and initial recognition works great.

**Setup:** Snapmaker U1 on the 1.5.2 pre-release, Spoolman 0.26 as backend, spools tagged and linked via SpoolKit (both card UIDs per spool stored in `card_uids`).

**What happens:**
1. Insert four tagged spools into the feeders — all four are recognised and shown correctly. 
2. Load the extruders one after another (T0…T3).
3. A few seconds after each load completes, that channel's spool info is cleared again (vendor/type/color back to NONE, spool id back to 0). After loading all four, only the last channel still shows its filament.

From moonraker.log (spool ids 8-11 are my four spools):

```
09:25:12 [spoollink] spool_ids changed: [9, 8, 10, 11]        <- all four resolved
09:26:01 [spoollink] toolhead extruder changed: extruder3 -> extruder
09:26:07 [spoollink] spool_ids changed: [9, 8, 10, 11] -> [0, 8, 10, 11]
09:26:57 [spoollink] toolhead extruder changed: extruder -> extruder1
09:27:02 [spoollink] spool_ids changed: [0, 8, 10, 11] -> [0, 0, 10, 11]
09:28:10 [spoollink] spool_ids changed: [0, 0, 10, 11] -> [0, 0, 0, 11]
09:29:17 [spoollink] spool_ids changed: [0, 0, 0, 11] -> [0, 0, 0, 0]
```

The card UID stays present and unchanged the whole time. My reading (please correct me): the firmware re-reads the card during the load, and since my spools' second tag carries no filament data (UID only — the factory-written tag doesn't always face the antenna), it writes empty display fields over what SpoolLink had applied. SpoolLink only re-resolves when the UID *changes*, so the cleared channel stays "unknown".

[...]

PR #649, opened 2026-08-18
feat(spoollink): (Mod) Add Spoolman Pressure Advance integration & optional auto-calibration on load

pechex:
## 📌 Summary

This PR introduces two modular, cleanly separated commits to enhance **Spoolman** integration for the Snapmaker U1:

1. **Spoolman Pressure Advance Sync (`Commit 1`)**:
   - Automatically saves measured Pressure Advance to Spoolman upon completion of flow calibration (`FLOW_CALIBRATE`).
   - Automatically creates the `pressure_advance` custom float field in Spoolman if missing.
   - Applies the saved `pressure_advance` from Spoolman directly to the active extruder stepper whenever a spool is loaded/resolved via SpoolLink.

2. **Optional Auto Pressure Advance Calibration on Load (`Commit 2`)**:
   - Adds an optional toggle to the `/firmware-config/` screen (`spoolman_auto_pa`) dependent on Spoolman being enabled.
   - Automatically queues and triggers `FLOW_CALIBRATE` on physical load completion (`AUTO_LOAD` → `IDLE`) if the loaded Spoolman spool lacks a saved Pressure Advance value.
   - Features a 2-second stabilization delay for multi-extruder loads, automatic toolhead switching (`T0`-`T3`), and a multi-layered anti-loop safety shield limiting calibration to 1 attempt per spool session.

imagen

---

### Commit 1: `feat(spoollink): Add Spoolman Pressure Advance integration (save on calibration, apply on load)`
- **`01-flow_calibrator.patch`**: Hooks into `flow_calibrator.py` to invoke `spoollink_save_pa` via RPC upon successful calibration. Uses `hasattr()` check so it functions standalone even if Commit 2 is not merged.
- **`01-spoollink.patch` (Moonraker)**: Ensures custom `pressure_advance` float field exists in Spoolman, registers `spoollink_save_pa` RPC, and passes `SPOOLMAN_PA` in `info` struct.
- **`02-spoollink.patch` (Klipper)**: Applies `SPOOLMAN_PA` to the extruder stepper upon spool resolution.

### Commit 2: `feat(spoollink): Add optional automatic Pressure Advance calibration on filament load`
- **`28_settings_spoolman_auto_pa.yaml`**: Web UI setting definition for `/firmware-config/` under Snapmaker components.
- **`02-spoollink.patch` (Klipper auto-calibration)**:
  - Listens to `machine_state_manager.post_hooks` for `LOAD` / `FLOW_CALIBRATION` → `IDLE` transitions.
  - 2-second delayed reactor timer to handle sequential multi-extruder loads smoothly.
  - Automatically switches active toolhead (`T<channel>`) if needed before running `FLOW_CALIBRATE`.
  - Multi-layer anti-loop shield limiting calibration attempts to 1 per spool insertion session.

---

## 🧪 Verification Plan

- [x] Tested manual flow calibration: PA value is saved to Spoolman spool `#id` and printed to console.
- [x] Tested loading spool with existing PA in Spoolman: PA is applied to stepper immediately, auto-calibration is skipped.
- [x] Tested loading spool without PA in Spoolman: Physical load completes, toolhead switches, auto-calibration runs and saves PA to Spoolman.
- [x] Tested multi-extruder sequential loads: All extruders load physically first, then auto-calibrate in sequence


shivalwolf:
This is a fantastic Idea. i absolutely love it.

I tried using the auto build, however the config menu options in the firmware page do not show up.

pechex:
> This is a fantastic Idea. i absolutely love it.
> 
> I tried using the auto build, however the config menu options in the firmware page do not show up.

Yes, you are right,. It was building the regular unmodded version. It should be OK now, please check again

image

https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/actions/runs/32203297772/job/95921405483#step:14:3271

shivalwolf:
It shows up now.
However doing my 1st attempt to load a filament results in some odd behaviour.

I have filament loaded in E0 E1 E3
The filaments in those Extruders do not have saved data.

E2 was unloaded. When starting a Load procedure for E2 the screen goes blank except for the X to close the loading.
However the printer starts the Flow calibration on E0, Continues to E1

if hitting X on the load screen the flow calibration continues and data is saved for the earlier filaments. 
When hitting load on E2 again the printer throws a Filament runout detected error as it has not yet pulled the filament through for the load procedure.


I believe the solution to this would be if there is no saved data for Extruders that were already loaded it should do the load procedure on the selected Extruder 1st and then continue onto the earlier ones after the load and calibration is complete for the specifically selected extruders

shivalwolf:
It appears that your firmware build contains this issue that's been around for a while when hitting the load button on the display [Issue 628](https://github.com/paxx12-snapmaker-u1/SnapmakerU1-Extended-Firmware/issues/628)

In between hitting the load button and it feeding the filament in the Spool ID mapping gets cleared. And while the auto calibration completes as expected after the load procedure because the mapping had been lost  so the pressure advance is not saved against the spool

if after manually loading you quickly go into the filament manager and click the refresh button, the mapping is returned and it gets saved

pechex:
> It shows up now. However doing my 1st attempt to load a filament results in some odd behaviour.
> 
> I have filament loaded in E0 E1 E3 The filaments in those Extruders do not have saved data.
> 
> E3 was unloaded. When starting a Load procedure for E2 the screen goes blank except for the X to close the loading. However the printer starts the Flow calibration on E0, Continues to E1
> 
> if hitting X on the load screen the flow calibration continues and data is saved for the earlier filaments. When hitting load on E2 again the printer throws a Filament runout detected error as it has not yet pulled the filament through for the load procedure.
> 
> I believe the solution to this would be if there is no saved data for Extruders that were already loaded it should do the load procedure on the selected Extruder 1st and then continue onto the earlier ones after the load and calibration is complete for the specifically selected extruders

Hi @shivalwolf thank you so much for testing this!!

I wasn't able to reproduce this. 

1. Disabled the auto-calibrate functionality. 
2. Loaded a spool with a tag and already created in Spoolman. 
3. Didn´t calibrate Auto PA as the function was disabled
4. Enabled the function again 
5. Loaded another spool with a tag and already created in Spoolman. 
6. It just calibrated the new one, the first one was left untouched. 

Can you attach some logs to analyze?

shivalwolf:
I made a typo. It was E2 that was unloaded and then starting load on E2

"E2 was unloaded. When starting a Load procedure for E2 the screen goes blank except for the X to close the loading.
However the printer starts the Flow calibration on E0, Continues to E1"


I Will try and collect some logs and see if i can reproduce the issue again myself and collect the logs from that (I am digging though yesterdays logs but can't find the time of the event)

shivalwolf:
I have been unable to repeat the issue again myself and I am not sure why.
My 1st thought is that the issue might have been just caused by a glitched being the 1st spools that were read after the firmware update. My 2nd thought is that it it only occurred the 1st time because the custom field didn't exist in spoolman yet.

It has been running cleanly the last few days with a number of spool swaps.

However the 2nd issue related to the tags clearing still occurs however there was a pull request last night that apparently resolves this issue in general in #657

shivalwolf:
Paxx12 made a comment that makes sense on the discord. that you might want to consider in this mod.
 — 18:35
storing PA, requires per-nozzle size (0.2, 0.4, 0.8) and per-nozzle type setting (standard vs high flow)


It could probably be done by storing the data as a string and making it store the type and size along with the particular reading. Or it could be done with having a new field created dynamically as it showed up or was needed. The code already creates the field if it doesn't exist so maybe that could be tweaked to include the size+type assuming that data is even obtainable during the loading procedure

pechex:
> I have been unable to repeat the issue again myself and I am not sure why. My 1st thought is that the issue might have been just caused by a glitched being the 1st spools that were read after the firmware update. My 2nd thought is that it it only occurred the 1st time because the custom field didn't exist in spoolman yet.
> 
> It has been running cleanly the last few days with a number of spool swaps.
> 
> However the 2nd issue related to the tags clearing still occurs however there was a pull request last night that apparently resolves this issue in general in #657

Hey, thanks so much for testing. It is a good point, maybe the error you saw happens when the field is not yet created. I will try to reproduce



> Paxx12 made a comment that makes sense on the discord. that you might want to consider in this mod. — 18:35 storing PA, requires per-nozzle size (0.2, 0.4, 0.8) and per-nozzle type setting (standard vs high flow)
> 
> It could probably be done by storing the data as a string and making it store the type and size along with the particular reading. Or it could be done with having a new field created dynamically as it showed up or was needed. The code already creates the field if it doesn't exist so maybe that could be tweaked to include the size+type assuming that data is even obtainable during the loading procedure

I will work on this. I think getting the nozzle size should be possible, but I'm not sure of the nozzle type. Do high flow nozzles even exist? My initial idea is to have a set up option in the firmware-config page for what type of nozzle you are currently using.

shivalwolf:
> I will work on this. I think getting the nozzle size should be possible, but I'm not sure of the nozzle type. Do high flow nozzles even exist? My initial idea is to have a set up option in the firmware-config page for what type of nozzle you are currently using.

High Flow from Snapmaker are in beta testing currently. They will become an option in the future.

pechex:
> > I will work on this. I think getting the nozzle size should be possible, but I'm not sure of the nozzle type. Do high flow nozzles even exist? My initial idea is to have a set up option in the firmware-config page for what type of nozzle you are currently using.
> 
> High Flow from Snapmaker are in beta testing currently. They will become an option in the future.

OK but as of now, is it known if there will be an option in the firmware to select the nozzle type? I believe the only change comes from the slicer, not really from the firmware. Maybe I should only add nozzle size and we can add nozzle type at a later point?

shivalwolf:
> OK but as of now, is it known if there will be an option in the firmware to select the nozzle type? I believe the only change comes from the slicer, not really from the firmware. Maybe I should only add nozzle size and we can add nozzle type at a later point?

From the Discord with the announcement for the 1.6.0 Beta
"Added support for 0.4mm High Flow Hotend. (Full hardware functionality also requires Snapmaker Orca V2.3.7 or later for proper configuration)"

So it sounds like that there is a firmware change that will be in place with 1.6.0 for the High Flow in the firmware.
However yeah its not in the main firmware yet so it may not be worth focusing on the the type yet and stick to the size

pechex:
@shivalwolf  I don´t suggest you install this one just yet, its buggy. I will update soon.

pechex:
It will now properly save multiple values of PA in a new field called pressure_advance_matrix.

{
"0.4_high_flow": 0.025524, 
"0.4_standard": 0.024544
}

it supports nozzle size and type.

shivalwolf:
> It will now properly save multiple values of PA in a new field called pressure_advance_matrix.
> 
> { "0.4_high_flow": 0.025524, "0.4_standard": 0.024544 }
> 
> it supports nozzle size and type.

Is this build ready to test out now. I unfortunately dont have other nozzles unfortunately. but i should be easily able to see if things load and unload and get the flow correctly.

pechex:
> > It will now properly save multiple values of PA in a new field called pressure_advance_matrix.
> > { "0.4_high_flow": 0.025524, "0.4_standard": 0.024544 }
> > it supports nozzle size and type.
> 
> Is this build ready to test out now. I unfortunately dont have other nozzles unfortunately. but i should be easily able to see if things load and unload and get the flow correctly.

Yes, you can test. You don't really need other nozzles, you can change from the printer screen and it will just "believe" you changed it. This is only for testing of course. I DO have 0.2 nozzles, but I didn't have the time to exchange them. WIl do it at some point,

shivalwolf:
So just updated.
As soon as the printer booted it started calibrating the currently loaded filimnets.
Not a problem just was not expected.

Its now done all of them and has the nice new values.

I also noticed that the old value was limited to 3 decimal places and the new matrix has 6 decimal places. more accuracy is useful

decay71:
Hi, multiACE maintainer here (Snapmaker U1 + Anycubic ACE). Nice work on this – I wanted to let you know that i have adopted your pressure_advance_matrix schema verbatim: same Spoolman extra field, same {diameter}_{volume_type} keys (e.g. 0.4_standard), raw Klipper values. I read and write it for omy own per-spool PA (values come from the stock FLOW_CALIBRATE and are applied when the spool is loaded), and in SpoolLink mode i leave PA entirely to your mod so the two never fight over the extruder.

The point is interop: a spool moving between a SpoolLink setup and a multiACE setup keeps its calibration because the field and keys are identical. So if you ever change the field name or key format, a short ping would be much appreciated - i'll follow. Thanks for the clean design!

PR #655, opened 2026-08-20
Harden SpoolLink same-UID recovery

Ali-Jab:
## Summary

Prevent SpoolLink assignments from being lost when a U1 RFID refresh or automatic filament load temporarily clears `filament_spool_id`.

The existing recovery handles the case where the UID remains unchanged. This update also handles home-tagged spools whose UID becomes empty while loading to the toolhead.

The fallback is deliberately narrow. It requires the same feeder to remain present with automatic loading enabled and a recent transition from the corresponding toolhead filament sensor.

No public APIs, configuration fields, or data formats change.

## Implementation

- Recognize same-UID recovery only when fresh reader or metadata evidence confirms a refresh.
- Cache the last successfully resolved UID and Spoolman payload for each channel in memory.
- Restore cached metadata when a qualified automatic load clears both the UID and spool ID.
- Send `CARD_UID=[]` during that restore rather than inventing or rebinding a tag.
- Preserve the active Spoolman assignment while a validated recovery is running.
- Permit only one restore or resolution task per channel.
- Reject stale asynchronous results.
- Cancel retained state after feeder removal, a different UID, an unqualified clear, or Klippy disconnect.
- Do not poll or retry the RFID reader.

## Relationship to #622

This addresses the same underlying problem as #622 but uses guarded recovery state rather than triggering from cleared vendor or material fields.

The two recovery mechanisms should not be combined because both could schedule recovery for the same refresh.

Dragon2203 reported repeated reads and resolutions with #622 during print-start auto-load. He also found that its vendor-based trigger could interact with `force_generic_vendor` from #619 and leave a lane unresolved as Generic.

[Dragon2203's comparative hardware report](https://github.com/Ali-Jab/SnapmakerU1-Extended-Firmware/commit/7374e4eaf81f32885a390c86c73c3a76840b9fc8#commitcomment-197031392)

### Ali

Initial idle validation:

- Channel 2, spool 27, UID `0C259367`.
- The channel spool ID changed `27 -> 0 -> 27`.
- The active Spoolman assignment remained 27.
- Recovery completed in approximately 1.18 seconds.
- The idle test did not change Spoolman usage fields.

Channel-3 automatic-load validation on 21 August 2026:

- Spool 28, UID `06A533EB`, recovered `28 -> 0 -> 28` in approximately 0.17 seconds.
- Spool 27, UID `0C259367`, recovered `27 -> 0 -> 27` in approximately 0.10 seconds.
- Each load caused one recovery.
- No repeated resolution occurred.
- The active Spoolman assignment did not become `None` during either recovery.
- Removing or replacing the spool cleared the previous assignment.
- A different UID did not inherit the previous spool.
- Spoolman remained connected with no pending usage reports.

The empty-UID fallback is covered by automated tests but was not triggered by Ali's RFID tags. External validation with a home-tagged spool is still required before this draft is marked ready.

### Dragon2203

- Tested print-start auto-load on two channels.
- Confirmed that the active assignment remained available to AFC and Home Assistant.
- Tested normal RFID detection, removal and reinsertion, and manual reassignment.
- Directly compared this implementation with #622.
- Tested a combined build containing #619 and #624 without observing conflicts. Those changes are not included in this PR.

### Automated checks

- 24 focused SpoolLink regression tests pass.
- Python bytecode compilation passes.
- `git diff --check` passes.
- All changed text files use LF line endings.
- The PR changes only `spoollink.py`, its test file, and the pull-request workflow.

## AI disclosure

The implementation and tests were developed with OpenAI Codex. Ali reviewed the changes and performed the physical printer validation.

### DIFF
--- a/overlays/firmware-extended/38-feature-spoollink/root/home/lava/moonraker/moonraker/components/spoollink.py
+++ b/overlays/firmware-extended/38-feature-spoollink/root/home/lava/moonraker/moonraker/components/spoollink.py
@@ -14,7 +14,7 @@ import asyncio
 import json
 import logging
 import os
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Set, Tuple
 
 if TYPE_CHECKING:
     from ..confighelper import ConfigHelper
@@ -55,6 +55,15 @@ class SpoolLink:
             url = "http://" + url
         self._spoolman_url = url
         self._cache_dir: Optional[str] = config.get("cache_dir", None)
+        configured_channels = config.getintlist(
+            "external_channels", [], separator=",")
+        invalid_channels = sorted({
+            ch for ch in configured_channels if ch < 0 or ch > 3})
+        if invalid_channels:
+            raise config.error(
+                "[spoollink] external_channels must contain only channel "
+                f"numbers from 0 to 3; invalid: {invalid_channels}")
+        self._external_channels: Set[int] = set(configured_channels)
         self.http_client: HttpClient = self.server.lookup_component("http_client")
         self.klippy_apis: APIComp = self.server.lookup_component("klippy_apis")
 
@@ -91,8 +100,10 @@ class SpoolLink:
 
     async def component_init(self) -> None:
         logging.info(
-            "spoollink starting (spoolman: %s, cache: %s)",
-            self._spoolman_url, self._cache_dir or "disabled")
+            "spoollink starting (spoolman: %s, cache: %s, "
+            "external channels: %s)",
+            self._spoolman_url, self._cache_dir or "disabled",
+            sorted(self._external_channels) or "none")
         await self._ensure_fields()
 
     # -- Klippy lifecycle ---------------------------------------------------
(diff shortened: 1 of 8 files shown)


overlays/firmware-extended/38-feature-spoollink/root/home/lava/moonraker/moonraker/components/spoollink.py
```diff
@@ -110,33 +180,599 @@ def _handle_status_update(self, status: Dict[str, Any], eventtime: float) -> Non
...
+    def _deadline_remaining(deadline: float, eventtime: float) -> str:
+        if deadline <= 0. or eventtime <= 0.:
+            return "none"
+        return f"{deadline - eventtime:+.3f}s"
+
+    @staticmethod
+    def _field_is_clear(value: Any) -> bool:
+        return value in (None, "", "NONE", "None")
+
+    def _combined_ptc_clear_spool_ids(
```
paxx12, line 338: Why this is needed? I think this is incorrect to have so much logic to detect if slot is cleared or not. This breaks a very simple contract that spool link has today: always resolve spool when event is received.

paxx12 (changes requested):

PR #684, opened 2026-09-03
fix(spoollink): add missing fallback for E-prefixed lanes in SET_SPOO…

Fibdorius:
### **Description**
When assigning a spool via the Web UI without AFC-Lite installed, the macro fails with `unknown LANE 'E0'`. The comment above the macro states that it should fall back to the lane name itself (`E0` -> `0`), but the logic to parse this was missing.

### **Changes**
* Added an `elif` condition to check if the lane starts with `E` and has a valid length.
* Extracts the integer following the `E` prefix to map it to the correct channel.
* Preserves priority of `CHANNEL` parameter and `AFC_lane` definitions.

### **Testing**
Tsted gcode macro locally
Spool assignment via Web UI now successfully maps `LANE=E0` to `channel=0` and reports to Spoolman without raising an exception.

### DIFF
--- a/overlays/firmware-extended/38-feature-spoollink/root/usr/local/share/spoollink/klipper.cfg
+++ b/overlays/firmware-extended/38-feature-spoollink/root/usr/local/share/spoollink/klipper.cfg
@@ -43,4 +43,3 @@ gcode:
         { action_call_remote_method("spoollink_resolve_spool",
             channel=channel, spool_id=spool_id, card_uid=card_uid) }
     {% endif %}
-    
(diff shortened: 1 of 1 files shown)


overlays/firmware-extended/38-feature-spoollink/root/usr/local/share/spoollink/klipper.cfg
```diff
@@ -39,4 +42,4 @@ gcode:
         {% endif %}
         { action_call_remote_method("spoollink_resolve_spool",
             channel=channel, spool_id=spool_id, card_uid=card_uid) }
-    {% endif %}
+    {% endif %}
```
paxx12: Please add empty line at the end of this file.

PR #727, opened 2026-09-16
Allow all sizes of high flow nozzle on GUI

nuclearmistake:
Codex patched the gui bin.

The M.O. of this change is that the gui would **_hide_** all HF size options but 0.4 when HF is selected under nozzle info.

The patch makes the conditional an "if false", so no sizes of HF are hidden.

image


nuclearmistake:
https://github.com/OrcaSlicer/OrcaSlicer/pull/15755

Testing to get the high flow mismatch warning to go away when highflow nozzle configured in gui (even 0.4mm) when slicing with HF nozzle in orcaslicer.

I assume snorca wouldn't get a warning when using HF profiles.

PR #728, opened 2026-09-17
Updating to OctoEverywhere 5.3.5

QuinnDamerell:
This PR does a few things:

- Updates to OctoEverywhere  `5.3.5`
   - I added more logic to better handle and surface U1-specific error codes and states to users.
- Adds support for authenticated Moonraker instances
   - I checked to ensure that `DisableMoonrakerConfigFileWrites` prevents any writes to the Moonraker config file.
- Adds a new log line in the OctoEverywhere logs that indicates whether the plugin is account-linked, so the installer/updater doesn't have to wait to grep the logs for 15 seconds.

I will run a validation test tomorrow, once the build for this PR finishes, and report back!


QuinnDamerell:
@paxx12 I confirmed the upgrade, removal, and clean install paths all work as expected!

QuinnDamerell:
Let me know if you need anything else from me :)

